PR #2, opened 2023-03-01
Feature/core UI composables and styles 

ArtemisSoftware:
Coure ui module that includes composables for Text. It also includes Style, colors, fonts


core-ui/src/main/java/com/artemissoftware/core_ui/text/TYText.kt
```diff
@@ -0,0 +1,71 @@
...
+        text = text,
+        modifier = modifier,
+        color = color,
+        textAlign = textAlign,
+        style = style,
+        letterSpacing = letterSpacing,
+        overflow = overflow,
+        maxLines = maxLines
+    )
+}
```
philipplackner, line 39: Isn't this mostly already the default arguments for a text composable? I'm not sure if I see the value in creating this `TYText` (also considering it's not clear why it's `TY` - is it for Tasky?).

philipplackner (approved):

core-ui/src/main/java/com/artemissoftware/core_ui/text/TYText.kt
ArtemisSoftware: In a big project I usually create this composable because it allows in case all the Text around the app have a custom behaviour I will only need to change in one place. 
An example of this is having a particular letterSpacing or if the default color is a specific one or even some kind of shimmer effect. 
This will allow me to right less code.

Regarding why I call it TYText its because it is specific for Tasky. 
I just took the first letter and the last. 
This way I can identifiy every composable that should be used instead of the default one.

This means that all developes in the project should use TYText instead of Text, TYCard instead of Card. 
This will allow for more cohesion all around

PR #3, opened 2023-03-01
Feature/core UI buttons

ArtemisSoftware:
Composables relatred to buttons, textfields and identifiers


core-ui/src/main/java/com/artemissoftware/core_ui/composables/button/TaskySquareButton.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import com.artemissoftware.core_ui.R
+import com.artemissoftware.core_ui.theme.Black
+import com.artemissoftware.core_ui.theme.White
+
+@Composable
+fun TYSquareButton(
```
MKiperszmid, line 21: Rename to TaskySquareButton

MKiperszmid, line 21: Rename to TaskySquareButton

core-ui/src/main/java/com/artemissoftware/core_ui/composables/button/TaskyTextButton.kt
```diff
@@ -0,0 +1,34 @@
...
+import androidx.compose.material.MaterialTheme
+import androidx.compose.material.TextButton
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.tooling.preview.Preview
+import com.artemissoftware.core_ui.composables.text.TaskyText
+import com.artemissoftware.core_ui.theme.Silver
+
+@Composable
+fun TaskyTextButton(
```
MKiperszmid, line 12: What's the visual difference between TaskyButton and TaskyTextButton?

core-ui/src/main/java/com/artemissoftware/core_ui/composables/textfield/TaskyOutlinedTextField.kt
```diff
@@ -0,0 +1,240 @@
...
+                    }
+                )
+            }
+        }
+    }
+}
+
+private fun getVisualTransformation(isPasswordVisible: Boolean) = if (isPasswordVisible) VisualTransformation.None else PasswordVisualTransformation()
+
+@SuppressLint("UnrememberedMutableState")
```
MKiperszmid, line 170: Why are you getting this warning?

core-ui/src/main/java/com/artemissoftware/core_ui/composables/textfield/TaskyTextFieldType.kt
```diff
@@ -0,0 +1,23 @@
...
+    @Composable
+    fun getKeyboardType(): KeyboardType = when (this) {
+        EMAIL -> KeyboardType.Email
+        PASSWORD -> KeyboardType.Password
+        else -> KeyboardType.Text
+    }
+
+    fun getMaxChar(): Int = when (this) {
+        EMAIL -> 28
+        else -> 32
```
MKiperszmid, line 21: Why do we have a max of 28 and 32?

core-ui/src/main/java/com/artemissoftware/core_ui/composables/textfield/TaskyOutlinedTextField.kt
```diff
@@ -0,0 +1,240 @@
...
+
+
+@SuppressLint("UnrememberedMutableState")
+@OptIn(ExperimentalFoundationApi::class)
+@Preview(showBackground = true)
+@Composable
+private fun TaskyOutlinedTextFieldPreview() {
+
+    var text by remember {
+        mutableStateOf("Test")
```
MKiperszmid, line 201: No need to have a state on the preview. You can simply hardcode values , and make the "onValueChange" empty

core-ui/src/main/java/com/artemissoftware/core_ui/composables/textfield/TaskyOutlinedTextField.kt
```diff
@@ -0,0 +1,240 @@
...
+            hint = "password",
+            validationState = TaskyTextFieldValidationStateType.INVALID
+        )
+
+        TaskyOutlinedTextField(
+            taskyTextFieldType = TaskyTextFieldType.EMAIL,
+            text = text,
+            onValueChange = {
+                text = it
+            },
```
MKiperszmid, line 236: ``` onValueChange = {} ```

MKiperszmid: So you can remove the text state above, which is not needed on a preview

core-ui/src/main/java/com/artemissoftware/core_ui/composables/textfield/TaskyOutlinedTextField.kt
```diff
@@ -0,0 +1,240 @@
...
+                    Icon(
+                        painter = painterResource(if (isPasswordVisible.value) R.drawable.ic_visibility else R.drawable.ic_visibility_off),
+                        contentDescription = ""
+                    )
+                }
+            )
+        }
+        else -> {
+
+            if (text.isNotBlank() && validationState == TaskyTextFieldValidationStateType.VALID) {
```
MKiperszmid, line 150: No need to ask "isNotBlank()", since the validation logic should already do that

core-ui/src/main/java/com/artemissoftware/core_ui/composables/textfield/TaskyOutlinedTextField.kt
```diff
@@ -0,0 +1,240 @@
...
+        },
+        shape = shape
+    )
+}
+
+
+@Composable
+private fun TrailingIcon(
+    taskyTextFieldType: TaskyTextFieldType,
+    onClick: (TextFieldValue) -> Unit,
```
MKiperszmid, line 127: What's the use of this onClick?

core-ui/src/main/java/com/artemissoftware/core_ui/composables/textfield/TaskyOutlinedTextField.kt
```diff
@@ -0,0 +1,240 @@
...
+                    )
+                }
+            )
+        }
+        else -> {
+
+            if (text.isNotBlank() && validationState == TaskyTextFieldValidationStateType.VALID) {
+                IconButton(
+                    onClick = {
+                        onClick.invoke(TextFieldValue(""))
```
MKiperszmid, line 153: Why are we doing this?
The check mark shouldn't do anything. So no need to have it as IconButton

core-ui/src/main/java/com/artemissoftware/core_ui/theme/TaskyColor.kt
```diff
@@ -13,3 +13,12 @@ val Light2 = Color(0xFFF2F3F7)
 val White = Color(0xFFFFFFFF)
 val Brown = Color(0xFF40492D)
 val Orange = Color(0xFFFDEFA8)
+
+/**
+ * A1A4B2
```
MKiperszmid, line 18: This comment isn't adding anything, since the color is the one from HintTextColor

Should be removed
Unless it's meaningful, in which case it should have a bit more description

MKiperszmid (approved):
Couple comments about some components

core-ui/src/main/java/com/artemissoftware/core_ui/composables/button/TaskyTextButton.kt
ArtemisSoftware: They allow to create 2 different type of buttons

Captura de ecrã 2023-03-01 175007

core-ui/src/main/java/com/artemissoftware/core_ui/composables/textfield/TaskyTextFieldType.kt
ArtemisSoftware: Removed this ans changed the logic

core-ui/src/main/java/com/artemissoftware/core_ui/composables/button/TaskyButton.kt
```diff
@@ -0,0 +1,50 @@
...
+    modifier: Modifier = Modifier,
+    text : String,
+    onClick: () -> Unit,
+    textColor: Color = White,
+    backGroundColor: Color = Black
+) {
+    Button(
+        modifier = modifier
+            .fillMaxWidth()
+            .height(55.dp),
```
philipplackner, line 29: I'd avoid using `fillMaxWidth/Height()` in a composable and only determine this at the screen level. Imagine, you'd want to show two of these buttons next to each other. Currently, you force each button to fill the whole width which would prevent this from being possible.
It's only okay if every component has exactly the same size (here hardcoding a height would be ok for example), but I'd avoid the rest 👍

core-ui/src/main/java/com/artemissoftware/core_ui/composables/button/TaskySquareButton.kt
```diff
@@ -0,0 +1,49 @@
...
+    modifier: Modifier = Modifier,
+    @DrawableRes icon: Int,
+    size: Dp = 56.dp,
+    iconColor: Color = White,
+    backGroundColor: Color = Black,
```
philipplackner, line 26: background is one word and therefore should be spelled `backgroundColor`

core-ui/src/main/java/com/artemissoftware/core_ui/composables/textfield/TaskyOutlinedTextField.kt
ArtemisSoftware: Removed

philipplackner (approved):

core-ui/src/main/java/com/artemissoftware/core_ui/theme/TaskyColor.kt
ArtemisSoftware: Removed

core-ui/src/main/java/com/artemissoftware/core_ui/composables/button/TaskySquareButton.kt
ArtemisSoftware: My mistake :(

core-ui/src/main/java/com/artemissoftware/core_ui/composables/button/TaskyButton.kt
ArtemisSoftware: Will refactor that

core-ui/src/main/java/com/artemissoftware/core_ui/composables/textfield/TaskyOutlinedTextField.kt
ArtemisSoftware: I did this to test the behaviour of onValueChange

core-ui/src/main/java/com/artemissoftware/core_ui/composables/button/TaskyTextButton.kt
MKiperszmid: Ahh gotcha. Sounds good

core-ui/src/main/java/com/artemissoftware/core_ui/composables/textfield/TaskyOutlinedTextField.kt
```diff
@@ -0,0 +1,240 @@
...
+    )
+}
+
+
+@Composable
+private fun TrailingIcon(
+    taskyTextFieldType: TaskyTextFieldType,
+    onClick: (TextFieldValue) -> Unit,
+    text: String,
+    isPasswordVisible: MutableState<Boolean>,
```
MKiperszmid, line 125: Don't send a mutable state, send it as Boolean
By sending a mutable state, you are making it so the composable TrailingIcon can change the visibility

That state is better to have it in the whole Textfield rather than the Icon

I would personally do it like this:

Have an onIconClick: () -> Unit
And on the textfield, change the state

ArtemisSoftware: Got it. Will make the change

MKiperszmid (approved):

core-ui/src/main/java/com/artemissoftware/core_ui/composables/textfield/TaskyOutlinedTextField.kt
```diff
@@ -0,0 +1,225 @@
...
+    imeAction: ImeAction = ImeAction.Next,
+    validationState: TaskyTextFieldValidationStateType = TaskyTextFieldValidationStateType.NOT_VALIDATED
+) {
+
+    val focusManager = LocalFocusManager.current
+    val relocation = remember { BringIntoViewRequester() }
+    val scope = rememberCoroutineScope()
+
+    val textFormatted = maxChar?.let {
+        text.take(maxChar)
```
MKiperszmid, line 56: "maxChar" in this line, could be replaced with "it"

``` text.take(maxChar) ```

MKiperszmid (approved):

PR #4, opened 2023-03-01
Feature/data

ArtemisSoftware:
Dtos for agenda and authentication


MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/data/remote/TaskyApi.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import com.artemissoftware.tasky.data.remote.dto.authentication.LoginBodyDto
+import com.artemissoftware.tasky.data.remote.dto.authentication.LoginResponseDto
+import com.artemissoftware.tasky.data.remote.dto.authentication.RegistrationBodyDto
+import okhttp3.ResponseBody
+import retrofit2.http.Body
+import retrofit2.http.GET
+import retrofit2.http.POST
+
+interface TaskyApi {
```
philipplackner, line 11: I personally like to separate APIs based on their context, so you can also give features individual API instances. What I mean is that you could have one API interface just for the auth feature called `AuthApi` which combines the functions you have here.
Then another API could be `AgendaApi` for the rest. That also avoids clutter.

app/src/main/java/com/artemissoftware/tasky/data/remote/dto/authentication/LoginBodyDto.kt
```diff
@@ -0,0 +1,11 @@
+package com.artemissoftware.tasky.data.remote.dto.authentication
+
+
+import com.google.gson.annotations.SerializedName
+
+data class LoginBodyDto(
+    @SerializedName("email")
```
philipplackner, line 7: I'd recommend to use anything but Gson for JSON parsing. Gson is a Java library and has its issues with things like nullability and default values in Kotlin (for example, it can assign null values to nullable fields). Especially for Retrofit it's very easy to swap it out with a Kotlin library like Moshi or Kotlin serialization 😄 

Also, `@SerializedName` isn't needed if the field name stays the same, I personally leave it away then.

PR #7, opened 2023-03-01
Feature/data api souce

ArtemisSoftware:
Package restructure and api handling


core/src/main/java/com/artemissoftware/core/data/remote/exceptions/NetworkErrors.kt
```diff
@@ -0,0 +1,8 @@
+package com.artemissoftware.core.data.remote.exceptions
+
+object NetworkErrors {
+
+    val UNKNOWN_HOST = 1 to "UnknownHostException"
+    val UNKNOWN_API = 2 to "An error has occurred"
+    val GENERIC_API_ERROR = 3 to "An error has occurred"
```
MKiperszmid, line 7: 2 errors have the same value. Is this intended?

Also, they can probably be moved to a string resources
https://www.youtube.com/watch?v=JnJ_CaEMeEQ

MKiperszmid (commented):
One small comment

core/build.gradle
```diff
@@ -48,4 +48,8 @@ dependencies {
     androidTestImplementation rootProject.tests.espressoCore
 
     implementation rootProject.libraries.composeFoundation
+
+    //Retrofit
+    implementation rootProject.libraries.retrofit
+    implementation rootProject.libraries.retrofitConverterGson
```
philipplackner, line 54: I'd recommend to use anything but Gson for JSON parsing. Gson is a Java library and has its issues with things like nullability and default values in Kotlin (for example, it can assign null values to nullable fields). Especially for Retrofit it's very easy to swap it out with a Kotlin library like Moshi or Kotlin serialization 😄

core/src/main/java/com/artemissoftware/core/data/remote/HandleApi.kt
```diff
@@ -0,0 +1,54 @@
...
+import java.net.UnknownHostException
+
+object HandleApi {
+
+    suspend fun <T> safeApiCall(callFunction: suspend () -> T): T {
+        return try {
+
+            callFunction.invoke()
+
+        } catch (ex: Exception) {
```
philipplackner, line 16: Be careful with catching general exceptions inside suspend functions. You can break coroutine cancellation that way. If a coroutine is cancelled, the currently suspending function (here, it would be `FUNCTION_NAME`) will throw a CancellationException. If your catch block eats that up, this exception won't be properly propagated up the coroutine scopes which can cause some coroutines to run longer than they should.
You can fix this by adding this:

```
if(e is CancellationException) throw e
```
or alternatively by only catching specific exceptions and not CancellationExceptions.

In this video I show the issue with that a bit more in detail (starting at 17:45):
https://www.youtube.com/watch?v=VWlwkqmTLHc

core/src/main/java/com/artemissoftware/core/data/remote/exceptions/NetworkErrors.kt
philipplackner: Yeah either string resources, or probably even better to just return error codes to the UI and let the UI decide what kind of message it maps them to. If it's only used for the exception and not for the UI you can hardcode it there

philipplackner (approved):

core/src/main/java/com/artemissoftware/core/data/remote/exceptions/NetworkErrors.kt
ArtemisSoftware: I did a refactor. I would let a generic message be passed. But on the UI I will do a different mapping of codes to message

philipplackner (approved):

PR #8, opened 2023-03-02
Feature/domain

ArtemisSoftware:
Validation usecases and tests


app/src/test/java/com/artemissoftware/tasky/authentication/domain/usecases/validation/ValidatePasswordUseCaseTest.kt
```diff
@@ -0,0 +1,56 @@
+package com.artemissoftware.tasky.authentication.domain.usecases.validation
+
+import org.junit.Assert
+import org.junit.Before
+import org.junit.Test
+
+class ValidatePasswordUseCaseTest {
```
MKiperszmid, line 7: Great you are adding Unit Tests 🔥

core/src/main/java/com/artemissoftware/core/util/extensions/StringExt.kt
```diff
@@ -0,0 +1,16 @@
+package com.artemissoftware.core.util.extensions
+
+fun String.asInitials(limit: Int = 2): String {
```
MKiperszmid, line 3: This method you'll probably just use it once

I personally would convert it to a UseCase, since using it as extension makes it so all strings can use the method
And in this case, you just want to use it for the Full Name initials

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/authentication/domain/usecases/validation/ValidatePasswordUseCase.kt
```diff
@@ -0,0 +1,20 @@
...
+        if (password.isEmpty()) {
+            return false
+        }
+
+        return PASSWORD_REGEX.toRegex().matches(password)
+    }
+
+    companion object{
+
+        private const val PASSWORD_REGEX = "^(?=.*[a-z])(?=.*[A-Z])(?=.*\\d)[a-zA-Z\\d]{$MIN_CHARACTERS_FOR_PASSWORD,}\$"
```
philipplackner, line 18: I would personally avoid regex if not really necessary. Nobody will be able to quickly tell you what this matches for. A function like this would be much more readable:
```
val hasValidLength = password.length in (MIN_CHARACTERS_FOR_PASSWORD..MAX_CHARACTERS_FOR_PASSWORD)
val hasUppercase = password.any { it.isUppercase() }
val hasLowercase = password.any { it.isLowercase() }
val hasDigits = password.any { it.isDigit() }

return hasValidLength && hasUppercase && hasLowercase && hasDigits
```

app/src/test/java/com/artemissoftware/tasky/authentication/domain/usecases/validation/ValidateUserNameUseCaseTest.kt
```diff
@@ -0,0 +1,52 @@
+package com.artemissoftware.tasky.authentication.domain.usecases.validation
+
+import org.junit.Assert
+import org.junit.Before
+import org.junit.Test
+
+class ValidateUserNameUseCaseTest {
```
philipplackner, line 7: Here I'm missing a test that checks if full names with middle names are accepted, like `First Middle Last`

core/src/main/java/com/artemissoftware/core/domain/usecase/GetUserNameInitialsUseCase.kt
```diff
@@ -0,0 +1,20 @@
+package com.artemissoftware.core.domain.usecase
+
+class GetUserNameInitialsUseCase {
```
philipplackner, line 3: If a use case doesn't need constructor arguments, it could also be an `object` to not need to create an instance

core/src/main/java/com/artemissoftware/core/domain/usecase/GetUserNameInitialsUseCase.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    operator fun invoke(name: String, limit: Int = 2): String {
+
+        val buffer = StringBuffer()
+        name.trim().split(" ").filter {
```
philipplackner, line 8: What would happen here, if the full name would for example be `Philipp   Lackner`? Would this crash because it would split based on the 3 spaces in the middle and then try to access the first char of an empty string?

core/src/test/java/com/artemissoftware/core/domain/usecase/GetUserNameInitialsUseCaseTest.kt
```diff
@@ -0,0 +1,31 @@
...
+        val initials = "BW"
+        Assert.assertEquals(initials, getUserNameInitialsUseCase(name))
+    }
+
+    @Test
+    fun `enter name with 1 words get first letter from first word`() {
+
+        val name = "Bruce"
+        val initials = "B"
+        Assert.assertEquals(initials, getUserNameInitialsUseCase(name))
```
philipplackner, line 29: As per requirements, the initials should be `BR` here

philipplackner (approved):

core/src/test/java/com/artemissoftware/core/domain/usecase/GetUserNameInitialsUseCaseTest.kt
ArtemisSoftware: Fixed

core/src/main/java/com/artemissoftware/core/domain/usecase/GetUserNameInitialsUseCase.kt
ArtemisSoftware: After adding this specific test I verified it does not crash and returns the right initials

app/src/main/java/com/artemissoftware/tasky/authentication/domain/usecases/validation/ValidatePasswordUseCase.kt
ArtemisSoftware: refactored

app/src/main/java/com/artemissoftware/tasky/authentication/domain/usecases/validation/ValidatePasswordUseCase.kt
```diff
@@ -0,0 +1,17 @@
...
+
+    operator fun invoke(password: String): Boolean  {
+
+        val hasValidLength = password.length >= MIN_CHARACTERS_FOR_PASSWORD
+        val hasUppercase = password.any { it.isUpperCase() }
+        val hasLowercase = password.any { it.isLowerCase() }
+        val hasDigits = password.any { it.isDigit() }
+        val hasSpecialCharacter = password.any { !it.isLetterOrDigit() }
+
+        return hasValidLength && hasUppercase && hasLowercase && hasDigits && !hasSpecialCharacter
```
philipplackner, line 15: Why do you check if the password has NOT any special characters?

ArtemisSoftware: If I dont do this check the user will be able to introduce a password like "Pass#word1".
This way I can garantee only numbers and letters

app/src/main/java/com/artemissoftware/tasky/authentication/domain/usecases/validation/ValidateUserNameUseCase.kt
```diff
@@ -0,0 +1,13 @@
+package com.artemissoftware.tasky.authentication.domain.usecases.validation
+
+import com.artemissoftware.tasky.BuildConfig.*
+
+class ValidateUserNameUseCase {
+
+    operator fun invoke(name: String): Boolean  {
+
+        val hasValidLength =  name.length in (MIN_CHARACTERS_FOR_NAME..MAX_CHARACTERS_FOR_NAME)
+        return hasValidLength && name.isNotBlank()
```
MKiperszmid, line 10: If hasValidLength is true, then it will never be blank, so you could just

```
return hasValidLength
```

MKiperszmid (approved):
Small comments

PR #9, opened 2023-03-02
Feature/base classes 

ArtemisSoftware:
Base classes for events and viewmodels


core/src/main/java/com/artemissoftware/core/presentation/viewmodel/TaskyViewModel.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.artemissoftware.core.presentation.events.TaskyEvents
+import com.artemissoftware.core.presentation.events.UiEvent
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.flow.receiveAsFlow
+import kotlinx.coroutines.launch
+
+abstract class TaskyViewModel <E: TaskyEvents> : ViewModel() {
```
philipplackner, line 11: I'd personally avoid creating such base classes, they limit you more than they help you. Also they violate the single responsibility principle, since you can't tell the purpose if you just look at the class name (and if they had one purpose, you could give them a clear name). 
Base classes couple all the child classes strongly to them which makes changes harder, especially if they do many things. In this case for example, what if you want to wait until a certain event was fully sent into the channel? (execute something after `_uiEvent.send(event)`) This base class makes this impossible because it forces events to be sent in a separate coroutine scope.
Also, I can feel that your plan was to extend this with states as well. In that case it would not really be possible to properly handle process death, since that's quite individual for each ViewModel

philipplackner (approved):

PR #10, opened 2023-03-02
Feature/login screen

ArtemisSoftware:
login sreen and state

### DIFF
--- a/app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginScreen.kt
+++ b/app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginScreen.kt
@@ -28,24 +28,7 @@ import com.artemissoftware.tasky.ui.theme.TaskyTheme
 
 @OptIn(ExperimentalFoundationApi::class)
 @Composable
-fun LoginScreen() {
-
-    /* TODO: uncomment when view model is ready
-    val state = viewModel.state.collectAsState()
-
-    BuildLoginScreen(
-        state = state.value,
-        events = viewModel::onTriggerEvent,
-        email = viewModel.email,
-        password = viewModel.password
-    )
-*/
-}
-
-
-@OptIn(ExperimentalFoundationApi::class)
-@Composable
-private fun BuildLoginScreen(
+fun LoginScreen(
     state: LoginState,
     email: String,
     password: String
@@ -149,6 +132,6 @@ private fun BuildLoginScreen(
 
 @Preview(showBackground = true)
 @Composable
-private fun BuildLoginScreenPreview() {
-    BuildLoginScreen(state = LoginState(), "email", "password")
+private fun LoginScreenPreview() {
+    LoginScreen(state = LoginState(), "email", "password")
 }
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginScreen.kt
```diff
@@ -0,0 +1,154 @@
...
+}
+
+
+@OptIn(ExperimentalFoundationApi::class)
+@Composable
+private fun BuildLoginScreen(
+    state: LoginState,
+    email: String,
+    password: String
+) {
```
philipplackner, line 52: What's the purpose of this build function? Why not use the `LoginScreen` directly and initialize the ViewModel in the nav root? `LoginScreen` can then simply accept the `LoginState`.

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginState.kt
```diff
@@ -0,0 +1,10 @@
+package com.artemissoftware.tasky.authentication.presentation.login
+
+import com.artemissoftware.core.presentation.composables.textfield.TaskyTextFieldValidationStateType
+import com.artemissoftware.core.presentation.composables.textfield.TaskyTextFieldValidationStateType.*
+
+data class LoginState(
+    val emailValidationStateType: TaskyTextFieldValidationStateType = NOT_VALIDATED,
+    val passwordValidationStateType: TaskyTextFieldValidationStateType = NOT_VALIDATED,
```
philipplackner, line 8: What are valid options here? Can't an email and a PW not only be valid or invalid? If so, why not just call this `isValidEmail` and make it a boolean?

philipplackner (approved):

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginScreen.kt
ArtemisSoftware: I usually do this because the moment I put the viewmodel I loose the ability to preview the composable on android studio.

Its a hack

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginState.kt
ArtemisSoftware: I created 3 states for it because if I do isValidEmail as boolean (and put default value as false) we will get the textfield in red (as if there was mistake) before the user filled anything. THe same thing eill happen if isValidEmail default value is false

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginScreen.kt
philipplackner: Yeah, you don't need your ViewModel at all in the `LoginScreen` composable. You can do something like this in your NavHost:
```
NavHost(...) 
    composable("login") {
        val viewModel = hiltViewModel<LoginViewModel>()
        val state by viewModel.state.collectAsState()
        LoginScreen(state)
    }
}
```

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginState.kt
philipplackner: Ah right, that makes sense

PR #11, opened 2023-03-03
Feature/api sources

ArtemisSoftware:


app/src/test/java/com/artemissoftware/tasky/util/FakeData.kt
```diff
@@ -0,0 +1,14 @@
+package com.artemissoftware.tasky.util
+
+import com.artemissoftware.tasky.authentication.data.remote.dto.LoginResponseDto
+
+object FakeData {
+
+    val loginResponseDto = LoginResponseDto(
```
MKiperszmid, line 7: Is this going to be used outside of the AuthenticationMapperTest?
If not, then there's no need to make it a different object

MKiperszmid (approved):
Really small comments

app/src/test/java/com/artemissoftware/tasky/util/FakeData.kt
ArtemisSoftware: I will use it to make api related tests

PR #12, opened 2023-03-03
Feature/register screen

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginScreen.kt
```diff
@@ -24,14 +24,18 @@ import com.artemissoftware.core.presentation.theme.Link
...
 import com.artemissoftware.tasky.R
+import com.artemissoftware.tasky.authentication.presentation.login.composables.LoginForm
+import com.artemissoftware.tasky.authentication.presentation.register.RegisterEvents
+import com.artemissoftware.tasky.authentication.presentation.register.composables.RegisterForm
 import com.artemissoftware.tasky.ui.theme.TaskyTheme
 
 @OptIn(ExperimentalFoundationApi::class)
 @Composable
 fun LoginScreen(
     state: LoginState,
```
MKiperszmid, line 35: You are sending both the state and also email and password
Either send just the state (Which has email and password) or only the variables

I personally send the variables, because If im not mistaken, if the state changes by any means (maybe a isLoading or whatever), then it will trigger an unnecesary recomposition
In this one, I might be wrong, so I'll tag @philipplackner just to be sure

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginScreen.kt
```diff
@@ -59,44 +63,22 @@ fun LoginScreen(
...
-                                    onClick = {
-                                        // TODO: add event
-                                    }
-                                )
-                            }
+                            LoginForm(
+                                modifier = Modifier.align(Alignment.TopCenter),
+                                email = email,
+                                onEmailValueChange = {
+                                    events.invoke(LoginEvents.ValidateEmail(email = it))
```
MKiperszmid, line 70: Optiona:
The "invoke" part isn't really necessary. Since the parameter in the composable is a Lambda

You could just
```
events(LoginEvents.ValidateEmail(email = it))
```

app/src/main/res/drawable/ic_arrow_left.xml
```diff
@@ -0,0 +1,13 @@
...
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+  <path
+      android:pathData="M15,5L8,12.5L15,20"
+      android:strokeLineJoin="round"
+      android:strokeWidth="2"
+      android:fillColor="#00000000"
+      android:strokeColor="#ffffff"
+      android:strokeLineCap="round"/>
```
MKiperszmid, line 12: Isn't this arrow the same as
Icons.Default.ArrowBack (Or a similar like that one)? https://fonts.google.com/icons?icon.query=left

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/composables/LoginForm.kt
```diff
@@ -0,0 +1,85 @@
...
+import com.artemissoftware.tasky.R
+
+@OptIn(ExperimentalFoundationApi::class)
+@Composable
+fun LoginForm(
+    email: String,
+    password: String,
+    onEmailValueChange: (String) ->Unit,
+    onPasswordValueChange: (String) ->Unit,
+    onLoginClick: () ->Unit,
```
MKiperszmid, line 27: Since this are all tied with the LoginEvent

You could replace all of this with:
```
onEvent: (LoginEvents) -> Unit
```

And then on the places you use it, you would do:

Instead of:
```
onValueChange = onEmailValueChange,
```

Use:
```
onValueChange = {onEvent(LoginEvents.ValidateEmail(it)},
```

So basically you can remove a ton of lambas being send to Composables, just by sending this onEvent lambda ;)

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginScreen.kt
MKiperszmid: I think this comment is worthless if you end up doing what I mentioned in the LoginForm comment 

Meaning, way less code :D

MKiperszmid (approved):
Left some comments

I only reviewed LoginForm and LoginScreen, but the same applies to RegisterForm and RegisterScreen

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/composables/LoginForm.kt
ArtemisSoftware: Refactored

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginScreen.kt
ArtemisSoftware: Refactored

app/src/main/res/drawable/ic_arrow_left.xml
ArtemisSoftware: Can I use icons from this library or should I use all the Icons provided on figma?

PR #15, opened 2023-03-06
Feature/datastore

ArtemisSoftware:


core/src/main/java/com/artemissoftware/core/domain/models/authentication/User.kt
```diff
@@ -1,7 +1,7 @@
 package com.artemissoftware.core.domain.models.authentication
 
 data class User(
-    val fullName: String,
-    val token: String,
-    val id: String
+    val fullName: String = "",
+    val token: String = "",
+    val id: String = ""
```
MKiperszmid, line 6: Why are you setting default values?

I think you would never want any of those to be empty, since that would mean the information is corrupted

For example, if the token is empty, then the user isn't really authenticated
But what if token is filled, but fullname or id isnt?
That's why I recommend storing everything from the api, without having default values

MKiperszmid (approved):

core/src/main/java/com/artemissoftware/core/domain/models/authentication/User.kt
ArtemisSoftware: On the serealizer I need to define a default value. My approach was to add default values.
If I dont do that how can I implement this class?

`object UserStoreSerializer : Serializer<User> {

    override val defaultValue: User
        get() = User()
`

ArtemisSoftware: If the default value is NULL then I believe I will never be able to update it on the 
UserStoreRepositoryImpl on line 11

ArtemisSoftware: Please advise

MKiperszmid: Ah gotcha

Let's have it as an empty User then

MKiperszmid (approved):

PR #16, opened 2023-03-06
Feature/database

ArtemisSoftware:


core/src/main/java/com/artemissoftware/core/data/database/entities/ReminderTimeEntity.kt
```diff
@@ -0,0 +1,12 @@
+package com.artemissoftware.core.data.database.entities
+
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+@Entity
+data class ReminderTimeEntity(
```
MKiperszmid, line 7: What will be the use of this entity?

MKiperszmid (commented):
Just one quetion

core/src/main/java/com/artemissoftware/core/data/database/entities/ReminderTimeEntity.kt
ArtemisSoftware: I want to have the time for the notifications on a table
I will change name of the table to NotificationWarningEntity so its clearer.

ArtemisSoftware: I will also changed the name of one field in order to make clearer

PR #18, opened 2023-03-08
Feature/agenda api domain reminder

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/AgendaItem.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import com.artemissoftware.core.domain.models.agenda.AgendaItemType
+
+data class AgendaItem(
+    val id: String? = null,
+    val title: String,
+    val description: String? = null,
+    val remindAt: Long,
+    val time: Long,
+    val type: AgendaItemType
```
MKiperszmid, line 11: What would happen if this is an Event or a Task, which have extra parameters?

core/src/main/java/com/artemissoftware/core/data/database/converters/AgendaItemTypeConverter.kt
```diff
@@ -0,0 +1,18 @@
...
+class AgendaItemTypeConverter {
+
+    @TypeConverter
+    fun fromAgendaItemType(type: AgendaItemType): String {
+        return type.name
+    }
+
+    @TypeConverter
+    fun toAgendaItemType(type: String): AgendaItemType {
+        return AgendaItemType.valueOf(type)
```
MKiperszmid, line 15: Are you going to use just 1 database table for all 3 types of AgendaItems?

I recommend having 1 table for each item.

Because else, you'll have a ton of empty parameters in the database (ex: Reminder with empty "isCompleted" since that's used for Task"

MKiperszmid (approved):

core/src/main/java/com/artemissoftware/core/domain/models/agenda/AgendaItemType.kt
```diff
@@ -0,0 +1,8 @@
+package com.artemissoftware.core.domain.models.agenda
+
+enum class AgendaItemType {
```
MKiperszmid, line 3: Is this being used?

core/src/main/java/com/artemissoftware/core/data/database/converters/AgendaItemTypeConverter.kt
```diff
@@ -0,0 +1,18 @@
+package com.artemissoftware.core.data.database.converters
+
+import androidx.room.TypeConverter
+import com.artemissoftware.core.domain.models.agenda.AgendaItemType
+
+class AgendaItemTypeConverter {
```
MKiperszmid, line 6: Since you are now creating one table for each agenda item
Is this converter being used?

PR #19, opened 2023-03-08
Feature/photo screen

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/photo/PhotoEvents.kt
```diff
@@ -0,0 +1,8 @@
+package com.artemissoftware.tasky.agenda.photo
+
+import com.artemissoftware.core.presentation.events.TaskyEvents
+
+sealed class PhotoEvents: TaskyEvents() {
+    object PopBackStack: PhotoEvents()
+    data class DeletePhoto(val photoId: String): PhotoEvents()
```
MKiperszmid, line 7: There should also be an AddPhoto, right?

app/src/main/java/com/artemissoftware/tasky/agenda/photo/PhotoScreen.kt
```diff
@@ -0,0 +1,99 @@
...
+                    events(PhotoEvents.PopBackStack)
+                },
+                backGroundColor = Black,
+                title = stringResource(id = R.string.photo),
+                toolbarActions = { color->
+                    TaskyToolBarAction(
+                        iconId = R.drawable.ic_trash,
+                        tint = color,
+                        onClicked = {
+                            events(PhotoEvents.DeletePhoto(""))
```
MKiperszmid, line 45: This "" is temporary, right? When this is the case, I personally like to add a // TODO
Since then you can find all the TODO's in Android Studio. Making it really easy to remember to change it
image

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/photo/PhotoScreen.kt
ArtemisSoftware: Thanks for the tip :)

app/src/main/java/com/artemissoftware/tasky/agenda/photo/PhotoEvents.kt
MKiperszmid: Oh you are right

I got confused since someone else added an option to add from here

You are good

PR #21, opened 2023-03-10
Feature/app navigation

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/navigation/Destination.kt
```diff
@@ -0,0 +1,13 @@
+package com.artemissoftware.tasky.navigation
+
+import com.artemissoftware.core.presentation.navigation.BaseDestination
+import com.artemissoftware.core.presentation.navigation.models.CustomNavigationArgument
+
+sealed class Destination(
+    route: String,
+    customArguments: List<CustomNavigationArgument> = emptyList(),
+) : BaseDestination(route = route, customArguments = customArguments) {
```
philipplackner, line 9: Anything with the name `Base` in it feels like a code smell to me. You can't tell what this class contains, unless you take a look in it. Here I don't think it's as big of an issue as it would be for a `BaseViewModel` or so, but maybe I'd rather use the name `Destination` instead of `BaseDestination` and rename the sealed class to something more specific like `TaskyScreenDestination`

app/src/main/java/com/artemissoftware/tasky/navigation/NavigationRoute.kt
```diff
@@ -0,0 +1,61 @@
...
+                onPopBackStack = {
+                    navController.popBackStack()
+                },
+                onNavigate =  { event->
+                    navController.navigate(event.route)
+                },
+            )
+        }
+    }
+
```
philipplackner, line 59: At this point, I find your approach a bit brittle. You're creating quite a lot of coupling here, since you assume lots of things:
1. Each screen HAS TO HAVE a ViewModel
2. Each screen HAS TO HAVE the same arguments (referring to function parameters here, not nav arguments)
3. Each screen HAS TO HAVE a Scaffold as root
4. Each screen HAS TO show errors as snackbar

While the intention to limit the duplicate code is good, I wouldn't put this over everything else and over gaining flexibility. 

Also, I would avoid coupling your screen composables with their ViewModel. That will cause that you'll have to provide a fake ViewModel for every preview. Since ViewModels often depend on lots of other dependencies your screen does not depend on, you'll have to create fakes for these as well for the preview.
Instead, I recommend this:
```
composable("login") {
    val viewModel = hiltViewModel<LoginViewModel>()
    val state: LoginState by viewModel.loginState.collectAsState()
    LoginScreen(
        state = state,
        onEvent = viewModel::onEvent
    )
}
```
where the ViewModel could look something like this:
```
@HiltViewModel
class LoginViewModel @Inject constructor(...): ViewModel() {

    fun onEvent(event: LoginEvent) {
        when(event) {
             is LoginEvent.EmailChanged -> ...
             is LoginEvent.Login -> ....
        }
    } 
}
```

core/src/main/java/com/artemissoftware/core/presentation/composables/navigation/ManageUIEvents.kt
```diff
@@ -0,0 +1,31 @@
...
+
+@Composable
+fun ManageUIEvents(
+    uiEvent: Flow<UiEvent>,
+    scaffoldState: TaskyScaffoldState,
+    onNavigate: (UiEvent.Navigate) -> Unit = {},
+    onPopBackStack: () -> Unit = {},
+) {
+
+    LaunchedEffect(key1 = Unit) {
```
philipplackner, line 19: Here, you'd also need to pass the scaffoldState, to make sure this `LaunchedEffect` always uses the most up-to-date state

core/src/main/java/com/artemissoftware/core/presentation/navigation/BaseDestination.kt
```diff
@@ -0,0 +1,53 @@
...
+        return if(customArguments.isEmpty()) route else routeWithArguments
+    }
+
+    private val routeWithArguments: String = buildString {
+        append(route)
+        customArguments.forEachIndexed { index, custom ->
+            val symbol = if (index == 0) "?" else "&"
+            append("$symbol${custom.key}={${custom.key}}")
+        }
+    }
```
philipplackner, line 22: How do you handle mandatory arguments here? With your approach, you can only use optional ones. A route could look like this for example:
```
navigate("event_detail/$eventId/$date?isEditable=true")
```

philipplackner (commented):
All in all, the idea is good, but I'd personally not use it like this in my project. It creates too much coupling as per my comments.
I recommend to take a look at the Compose nav destinations library: https://github.com/raamcosta/compose-destinations

While I understand that some people might not want to use a 3rd party lib for navigation, it's a well tested library and takes away all that pain you had with your code here. All you then need is this:
```
@Destination
@Composable
fun LoginScreen(...) { }
```
And it will generate type-safe arguments for your screens

app/src/main/java/com/artemissoftware/tasky/navigation/Destination.kt
ArtemisSoftware: Agreed

core/src/main/java/com/artemissoftware/core/presentation/navigation/BaseDestination.kt
ArtemisSoftware: I overlooked this situation.

But I could create an a function for each route where I specify the arguments it should receive and of those wich ones are mandatory. And this function creates the route with arguments

On the viewmodels I would get this arguments since hilt allows this.

Since I am defining the navigation cant I just decide what is allowed or not?

ArtemisSoftware:
> All in all, the idea is good, but I'd personally not use it like this in my project. It creates too much coupling as per my comments. I recommend to take a look at the Compose nav destinations library: https://github.com/raamcosta/compose-destinations
> 
> While I understand that some people might not want to use a 3rd party lib for navigation, it's a well tested library and takes away all that pain you had with your code here. All you then need is this:
> 
> ```
> @Destination
> @Composable
> fun LoginScreen(...) { }
> ```
> 
> And it will generate type-safe arguments for your screens



> All in all, the idea is good, but I'd personally not use it like this in my project. It creates too much coupling as per my comments. I recommend to take a look at the Compose nav destinations library: https://github.com/raamcosta/compose-destinations
> 
> While I understand that some people might not want to use a 3rd party lib for navigation, it's a well tested library and takes away all that pain you had with your code here. All you then need is this:
> 
> ```
> @Destination
> @Composable
> fun LoginScreen(...) { }
> ```
> 
> And it will generate type-safe arguments for your screens

I will give this library a chance. 
But a question, in a situation where I am on screen A, I navigate to screen B fill the telephone field, and then return to screen A where I need the telephone I filled on screen B can I achieve this?

Another questiont does this library scale well with bigger projects where you have deeplinks and push notifications?

MKiperszmid (approved):

PR #22, opened 2023-03-10
Feature/login usecases

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/authentication/domain/usecases/LoginUseCase.kt
```diff
@@ -0,0 +1,38 @@
...
+            is ApiNetworkResponse.Error -> {
+                val error = result.exception?.description ?: UiText.DynamicString(LOGIN_ERROR)
+                emit(Resource.Error(error))
+            }
+            is ApiNetworkResponse.Success -> {
+                result.data?.let { userStoreRepository.saveUser(it) }
+                emit(Resource.Success(true))
+            }
+        }
+    }
```
philipplackner, line 32: For such network calls, I'd rather not use flows, that adds a bit much unnecessary complexity. In the end such a call only returns one thing: The data retrieved or the error.
Therefore, I'd rather make it just a suspend function.

Regarding loading, I'd handle that in the ViewModel, somehow like this:
```
viewModelScope.launch {
    isLoading = true
    login(...)
    isLoading = false
}
```

app/src/main/java/com/artemissoftware/tasky/authentication/domain/usecases/RegisterUserUseCase.kt
```diff
@@ -0,0 +1,36 @@
...
+            }
+            is ApiNetworkResponse.Success -> {
+                emit(Resource.Success(true))
+            }
+        }
+    }
+
+
+    companion object{
+        private const val REGISTER_ERROR = "An error occurred while doing registering"
```
philipplackner, line 34: I personally only make things a constant if that either makes them more readable or I use them at multiple places. Here, it's very obvious what this string is used for. It's also only used in this use case.

app/src/test/java/com/artemissoftware/tasky/authentication/domain/usecases/RegisterUserUseCaseTest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class RegisterUserUseCaseTest: BaseUseCaseTest() {
+
+    private lateinit var registerUserUseCase: RegisterUserUseCase
+    private lateinit var authenticationRepository: AuthenticationRepository
+
+    @Before
+    fun setUp() {
+        authenticationRepository = mock()
```
philipplackner, line 28: From my personal experience, mocks should be avoided as much as you can. It's okay to use them to test legacy code or code you don't own (from a library), but for code like this, better use fakes. There, you can be sure that every function always has a real return value and you're forced to extend the fake whenever your repository extends (since it implements the interface). For mocks, you're not forced to do that, which very often leads to breaking tests not because your app contains broken logic, but because of the mock.

philipplackner (approved):

app/src/main/java/com/artemissoftware/tasky/authentication/domain/usecases/LoginUseCase.kt
```diff
@@ -0,0 +1,31 @@
...
+    private val authenticationRepository: AuthenticationRepository
+){
+
+    suspend operator fun invoke(email: String, password: String): Resource<Boolean> {
+
+        val result = authenticationRepository.loginUser(email = email, password = password)
+
+        return when(result){
+            is ApiNetworkResponse.Error -> {
+                val error = result.exception?.description ?: UiText.DynamicString("An error occurred while doing login in")
```
MKiperszmid, line 22: Personal opinion

I'm not a fan of having error codes or strings that are going to be used on the app, in the domain
Imo, the approach would be to return an error (Maybe a custom error or a sealed class) and then map that error to a string on the presentation

MKiperszmid (approved):

app/src/test/java/com/artemissoftware/tasky/authentication/domain/usecases/LoginUseCaseTest.kt
```diff
@@ -0,0 +1,55 @@
...
+        val password = "Iambatman123"
+
+        val result = loginUseCase(email = email, password = password)
+
+        assert(result is Resource.Success)
+    }
+
+
+    @Test
+    fun `Login user with failure`() = runTest {
```
MKiperszmid, line 44: Here you could add 2 tests for failure
1- Test if the error message by the backend is returned 
2- Test if there's no error message by the backend, so you return your custom error

You add another variable to the FakeRepository like errorWithBackendMessage and the message just being "Backend error"
And you test on that

app/src/test/java/com/artemissoftware/tasky/authentication/domain/usecases/LoginUseCaseTest.kt
```diff
@@ -0,0 +1,55 @@
...
+import org.junit.Test
+import org.mockito.kotlin.mock
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class LoginUseCaseTest: BaseUseCaseTest() {
+
+    private lateinit var loginUseCase: LoginUseCase
+
+    private lateinit var userStoreRepository: UserStoreRepository
+    private lateinit var authenticationRepository: AuthenticationRepository
```
MKiperszmid, line 21: Here you could just "FakeUserStoreRepository" and "FakeAuthenticationRepository"

Since you want to always use fakes, no need to use the interface

Mostly to avoid downcasting like you did on line 46

app/src/test/java/com/artemissoftware/tasky/authentication/domain/usecases/RegisterUserUseCaseTest.kt
```diff
@@ -0,0 +1,57 @@
...
+        val fullName = "Bruce Wayne"
+
+        val result = registerUserUseCase(email = email, password = password, fullName = fullName)
+
+        assert(result is Resource.Success)
+    }
+
+
+    @Test
+    fun `Register user with failure`() = runTest {
```
MKiperszmid, line 45: Same about what I talked on Login test

MKiperszmid (approved):

MKiperszmid (approved):

PR #23, opened 2023-03-10
SplashScreen

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/authentication/presentation/splash/SplashScreen.kt
```diff
@@ -0,0 +1,43 @@
...
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.artemissoftware.core.R
+import com.artemissoftware.core.presentation.composables.icon.TaskyIcon
+import com.artemissoftware.core.presentation.composables.scaffold.TaskyScaffold
+import com.artemissoftware.core.presentation.theme.Green
+import com.artemissoftware.core.presentation.theme.White
+
+@Composable
+fun SplashScreen() {
```
MKiperszmid, line 17: Why not use Android's SplashScreen library?

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/splash/SplashScreen.kt
ArtemisSoftware: I just want it to be as flexible as possible.
So I can have a viewmodel to check the user state so I can decide if he should be redirected to registe, login or agenda

MKiperszmid: You can easily do that with the splashscreen api
Check Philipp's video ;)

PR #24, opened 2023-03-13
Feature/reminder usecases

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/AgendaRepositoryImpl.kt
```diff
@@ -23,16 +27,46 @@ class AgendaRepositoryImpl constructor(
...
+                ApiNetworkResponse.Success(data = true)
+
+            } catch (ex: TaskyNetworkException) {
+                ApiNetworkResponse.Error(exception = ex)
+            }
+        }
+    }
+    override suspend fun syncDelete(reminder: AgendaItem.Reminder): ApiNetworkResponse<Boolean> {
+
+        with(agendaApiSource) {
```
philipplackner, line 56: This `with` isn't needed here, you only need `agendaApiSource` once in this function. IMO, the indentations rather make it a bit harder to read compared to just writing `agendaApiSource.deleteReminder()`

app/src/main/java/com/artemissoftware/tasky/agenda/domain/usecase/reminder/DeleteReminderUseCase.kt
```diff
@@ -0,0 +1,27 @@
+package com.artemissoftware.tasky.agenda.domain.usecase.reminder
+
+import com.artemissoftware.core.data.SyncType
```
philipplackner, line 3: `SyncType` comes from data which you aren't allowed to use in domain. You can put this enum in domain instead :)

philipplackner (commented):
I personally find the caching logic a bit confusing as per my comments (mainly cause of the function names I think). What I like is your `SyncType` enum. In my implementation I did it a bit differently though:
1. I have a table `ModifiedReminderEntity` which saves the ID and sync type of a reminder
2. If a reminder is deleted, I insert the reminder's ID together with SyncType.DELETED in that table and immediately delete it from the DB
3. If a reminder is locally created/updated, but uploading it to the API failed, I insert its ID with SyncType.UPDATED/CREATED
4. I then run a WorkManager worker every 30min which goes through all modified agenda item tables and syncs them with the API

app/src/main/java/com/artemissoftware/tasky/agenda/data/mappers/AgendaMapper.kt
```diff
@@ -14,23 +17,37 @@ fun ReminderDto.toEntity(): ReminderEntity {
...
+
+fun AgendaItem.Reminder.toEntity(): ReminderEntity {
 
     return ReminderEntity(
         title = title,
         description = description,
-        id = this.id ?: id,
-        remindAt = remindAt,
-        time = time
+        id = this.id!!,
```
MKiperszmid, line 38: Why would Reminder have a nullable id tho?

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/ReminderReminderImpl.kt
```diff
@@ -0,0 +1,73 @@
...
+import com.artemissoftware.tasky.agenda.data.remote.source.AgendaApiSource
+import com.artemissoftware.tasky.agenda.domain.models.AgendaItem
+import com.artemissoftware.tasky.agenda.domain.repositories.ReminderRepository
+
+class ReminderReminderImpl constructor(
+    private val taskyDatabase: TaskyDatabase,
+    private val agendaApiSource: AgendaApiSource
+): ReminderRepository {
+
+    private val reminderDao = taskyDatabase.reminderDao
```
MKiperszmid, line 20: I prefer passing the DAO to the constructor rather than the db

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/ReminderReminderImpl.kt
```diff
@@ -0,0 +1,73 @@
...
+
+        return try {
+            when(syncType){
+                SyncType.CREATED -> {
+                    agendaApiSource.createReminder(reminder.toDto())
+                }
+                SyncType.UPDATED -> {
+                    agendaApiSource.updateReminder(reminder.toDto())
+                }
+                else ->{}
```
MKiperszmid, line 47: else -> Unit

It's a bit clearer than just leaving {}

And they both do the same

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/ReminderReminderImpl.kt
```diff
@@ -0,0 +1,73 @@
...
+            }
+            reminderDao.upsertReminderSync(ReminderSyncEntity(id = reminder.id!!, syncType = SyncType.SYNCED))
```
MKiperszmid, line 49: You are using !! too much on ids
Maybe if we have the id to not be nullable it should fix that

+ there's never a case where the id can be null

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/Notification.kt
```diff
@@ -0,0 +1,6 @@
+package com.artemissoftware.tasky.agenda.domain.models
+
+data class Notification(
```
MKiperszmid, line 3: Where are you using this class?

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/ReminderReminderImpl.kt
```diff
@@ -0,0 +1,73 @@
...
+        } catch (ex: TaskyNetworkException) {
+            ApiNetworkResponse.Error(exception = ex)
+        }
+    }
+
+    override suspend fun deleteAndUpdateSyncState(id: String) {
+        reminderDao.upsertSyncStateAndDelete(id = id, ReminderSyncEntity(id = id, syncType = SyncType.DELETED))
+    }
+
+    override suspend fun syncDelete(id: String): ApiNetworkResponse<Boolean> {
```
MKiperszmid, line 60: Are these 2 methods always called together? (DeleteReminderUseCase)

If so, then it can be just 1 method

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/ReminderReminderImpl.kt
ArtemisSoftware: Its just to make the code easier to read for the usecases

In the SaveReminderUseCase I did the same thing

app/src/main/java/com/artemissoftware/tasky/agenda/data/mappers/AgendaMapper.kt
philipplackner: Yeah as in WhatsApp, I also still don't understand this. Once you generate a random ID, it never needs to change

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/ReminderRepositoryImpl.kt
```diff
@@ -0,0 +1,71 @@
...
+import com.artemissoftware.tasky.agenda.domain.repositories.ReminderRepository
+
+class ReminderRepositoryImpl constructor(
+    private val reminderDao: ReminderDao,
+    private val agendaApiSource: AgendaApiSource
+): ReminderRepository {
+
+
+    override suspend fun getReminder(id: String): AgendaItem.Reminder {
+        return reminderDao.getReminderAndSyncState(id).toAgendaItem()
```
philipplackner, line 22: What if the ID doesn't exist for some reason?

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/ReminderRepositoryImpl.kt
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    override suspend fun save(reminder: AgendaItem.Reminder, syncType: SyncType) {
+
```
philipplackner, line 26: Is there a reason you often leave the first line of a function blank, but not always?

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/ReminderRepositoryImpl.kt
```diff
@@ -0,0 +1,71 @@
...
+                }
+                SyncType.UPDATED -> {
+                    agendaApiSource.updateReminder(reminder.toDto())
+                }
+                else -> Unit
+            }
+            reminderDao.upsertReminderSync(ReminderSyncEntity(id = reminder.id!!, syncType = SyncType.SYNCED))
+            ApiNetworkResponse.Success(data = true)
+        } catch (ex: TaskyNetworkException) {
+            ApiNetworkResponse.Error(exception = ex)
```
philipplackner, line 50: Does this exception also cover `IOException`? For example when there's no internet connection?

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/ReminderRepositoryImpl.kt
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    override suspend fun deleteAndUpdateSyncState(id: String) {
+        reminderDao.upsertSyncStateAndDelete(id = id, ReminderSyncEntity(id = id, syncType = SyncType.DELETED))
+    }
+
+    override suspend fun syncDelete(id: String): ApiNetworkResponse<Boolean> {
```
philipplackner, line 58: If there's no data to attach, I'd prefer `ApiNetworkResponse<Unit>` over `Boolean` since booleans are actually used for something while `Unit` is not

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/AgendaItem.kt
```diff
@@ -1,25 +1,31 @@
...
+
 sealed class AgendaItem(
     val itemId: String? = null,
     val itemTitle: String,
     val itemDescription: String? = null,
-    val itemRemindAt: Long,
-    val itemTime: Long
+    val itemRemindAt: LocalDateTime,
+    val itemTime: LocalDateTime,
+    val itemSyncState: SyncType
```
philipplackner, line 12: I'd maybe swap the names here, since you'll probably use these fields on a general `AgendaItem` more often than a casted one. What I mean is that I'd probably use `id`, `title`, etc. for these fields and then give each specific agenda item more specific fields like `reminderId`, `reminderTitle` and so on. But just my personal opinion, feel free to leave it

app/src/main/java/com/artemissoftware/tasky/agenda/domain/usecase/reminder/SaveReminderUseCase.kt
```diff
@@ -0,0 +1,34 @@
...
+        val result = reminderRepository.syncReminder(reminder = reminder, syncType = syncType)
+
+        when(result){
+            is ApiNetworkResponse.Error -> {
+                // TODO: should send message to the ui saying the sync failed?
+            }
+            is ApiNetworkResponse.Success -> {
+                // TODO: should send message to the ui saying the everything went well?
+            }
+        }
```
philipplackner, line 32: Could be an extension, but not needed here

app/src/test/java/com/artemissoftware/tasky/agenda/data/FakeReminderRepository.kt
```diff
@@ -0,0 +1,47 @@
...
+import com.artemissoftware.tasky.util.FakeData
+
+class FakeReminderRepository : ReminderRepository{
+
+    var returnNetworkError = false
+    private var reminders = mutableListOf(FakeData.reminder)
+
+    override suspend fun getReminder(id: String): AgendaItem.Reminder {
+        return reminders[0]
+    }
```
philipplackner, line 17: This doesn't simulate getting a reminder, since no matter which ID you pass, you won't get the reminder with that ID. Instead:
```
return reminders.find { it.id == id }
```

app/src/test/java/com/artemissoftware/tasky/agenda/data/FakeReminderRepository.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    override suspend fun deleteReminder(id: String) {
+        reminders.removeAt(0)
+    }
```
philipplackner, line 21: Same here, this should be:
```
val reminderToDelete = reminders.find { it.id == id } ?: return
reminders.removeAt(reminders.indexOf(reminderToDelete))
```

core/src/main/java/com/artemissoftware/core/data/database/TaskyDatabase.kt
```diff
@@ -2,18 +2,22 @@ package com.artemissoftware.core.data.database
...
 import com.artemissoftware.core.data.database.dao.NotificationWarningDao
+import com.artemissoftware.core.data.database.dao.ReminderDao
 import com.artemissoftware.core.data.database.entities.NotificationWarningEntity
 import com.artemissoftware.core.data.database.entities.ReminderEntity
+import com.artemissoftware.core.data.database.entities.ReminderSyncEntity
 
 @Database(
-    entities = [NotificationWarningEntity::class, ReminderEntity::class],
+    entities = [
+        NotificationWarningEntity::class,
```
philipplackner, line 13: What's that used for?

philipplackner (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/Notification.kt
MKiperszmid: Gotcha.

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/ReminderReminderImpl.kt
MKiperszmid: Mmm but you are calling both methods in the same use case
It makes more sense to just make it 1 method, and if in the future you need them split, then split it into 2
But right now, it makes more sense to just have it as 1

core/src/main/java/com/artemissoftware/core/data/database/TaskyDatabase.kt
ArtemisSoftware: It will contain all the notification

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/AgendaItem.kt
ArtemisSoftware: Noted

app/src/main/java/com/artemissoftware/tasky/agenda/domain/usecase/reminder/SaveReminderUseCase.kt
ArtemisSoftware: Noted

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/ReminderRepositoryImpl.kt
ArtemisSoftware: No reason in particular

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/ReminderRepositoryImpl.kt
ArtemisSoftware: I just added those conditions

MKiperszmid (approved):

PR #26, opened 2023-03-15
Feature/drop down menu

ArtemisSoftware:
Captura de ecrã 2023-03-15 180626
Captura de ecrã 2023-03-15 180504

### DIFF
--- a/core/src/main/java/com/artemissoftware/core/presentation/composables/DropDownMenu.kt
+++ b/core/src/main/java/com/artemissoftware/core/presentation/composables/DropDownMenu.kt
@@ -1,0 +1,145 @@
+package com.artemissoftware.core.presentation.composables
+
+import androidx.compose.animation.core.animateFloatAsState
+import androidx.compose.foundation.background
+import androidx.compose.foundation.border
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.*
+import androidx.compose.material.*
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.ArrowDropDown
+import androidx.compose.runtime.*
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.alpha
+import androidx.compose.ui.draw.rotate
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.layout.onGloballyPositioned
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.SpanStyle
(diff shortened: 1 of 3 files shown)


MKiperszmid (approved):

core/src/main/java/com/artemissoftware/core/presentation/composables/DropDownMenu.kt
```diff
@@ -0,0 +1,145 @@
...
+            allCaps = allCaps,
+            color = textColor,
+            style = textStyle
+        )
+
+        TaskyIcon(
+            size = 24.dp,
+            modifier = Modifier
+                .alpha(ContentAlpha.medium)
+                .rotate(degrees = angle)
```
philipplackner, line 79: For such transformations, I'd use the `graphicsLayer` modifier which doesn't cause recompositions AFAIK. If you pass the animated angle to your `rotate()` modifier, the `TaskyIcon` will recompose on every new angle. If you do it like this, it shouldn't do that since `graphicsLayer` is just called for every frame and independent of state:
```
TaskyIcon(
    modifier = Modifier.graphicsLayer {
        rotationX = angle
    }
)
```

core/src/main/java/com/artemissoftware/core/presentation/composables/DropDownMenu.kt
```diff
@@ -0,0 +1,145 @@
...
+
+        DropdownMenu(
+            modifier = Modifier
+                .width(with(LocalDensity.current) { parentSize.width.toDp() }),
+            expanded = expanded,
+            onDismissRequest = { expanded = false }
+        ) {
+            options.forEach { priority ->
+                DropdownMenuItem(
+                    modifier = Modifier.height(44.dp).padding(horizontal = 0.dp),
```
philipplackner, line 93: The `DropdownMenuItem` already makes sure the recommended size and padding is used, so this should be unnecessary

philipplackner (approved):

core/src/main/java/com/artemissoftware/core/presentation/composables/DropDownMenu.kt
```diff
@@ -0,0 +1,156 @@
...
+
+        DropdownMenu(
+            modifier = Modifier
+                .width(with(LocalDensity.current) { parentSize.width.toDp() }),
+            expanded = expanded,
+            onDismissRequest = { expanded = false }
+        ) {
+            options.forEachIndexed { index, item ->
+
+                val isLast = (index == options.size -1)
```
philipplackner, line 90: Little fun fact, Kotlin allows you to do this: `val isLast = index == options.lastIndex` 😁

PR #27, opened 2023-03-15
Feature/login viewmodel

ArtemisSoftware:


core/src/main/java/com/artemissoftware/core/presentation/composables/dialog/TaskyDialog.kt
```diff
@@ -239,19 +240,19 @@ private fun DialogOptions(
...
     val dialogTypeSuccess = TaskyDialogType.Success(
-        title = "Get updates",
-        description = "Allow permission to send notifications every day of the year",
+        title = UiText.DynamicString("Get updates"),
+        description = UiText.DynamicString("Allow permission to send notifications every day of the year"),
         dialogOptions = TaskyDialogOptions.DoubleOption(
-            confirmationText = R.string.ok,
-            cancelText = R.string.cancel
+            confirmationText = UiText.StringResource(R.string.ok),
+            cancelText = UiText.StringResource(R.string.cancel)
```
MKiperszmid, line 247: Why are some using stringresources and other just plain text?

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginViewModel.kt
```diff
@@ -0,0 +1,112 @@
...
+                title = UiText.StringResource(R.string.log_in),
+                description = ex.toUiText(),
+                dialogOptions = TaskyDialogOptions.DoubleOption(
+                    confirmationText = UiText.StringResource(R.string.retry),
+                    confirmation = {
+                        reloadEvent.invoke()
+                    },
+                    cancelText = UiText.StringResource(CoreR.string.cancel)
+                )
+            )
```
MKiperszmid, line 106: You are creating UI inside a ViewModel

UI should be created in UI
ViewModel simply handles the logic

It shouldn't create UI

MKiperszmid (approved):

core/src/main/java/com/artemissoftware/core/presentation/composables/dialog/TaskyDialog.kt
ArtemisSoftware: Before I would use some things as string resources and others as plain string.
Now I make everything a UiText. This also proofs how good UiText works :)

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginScreen.kt
```diff
@@ -13,33 +12,48 @@ import androidx.compose.ui.text.withStyle
...
-import com.artemissoftware.tasky.ui.theme.TaskyTheme
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flowOf
 
-@OptIn(ExperimentalFoundationApi::class)
 @Composable
-fun LoginScreen(
+fun LoginScreen(viewModel: LoginViewModel /* TODO : init viewmodel with Hilt when dependency is included on the project */) {
+
+    val state = viewModel.state.collectAsState().value
```
philipplackner, line 31: Instead of `.value`, this also works and looks a bit nicer: 
```
val state by viewModel.state.collectAsState()
```
Also, recently the `collectAsStateWithLifecycle()` dependency got stable, so I'd recommend to use that since it's the safest way to collect flows in Android (currently, your flow will be collected even if the app is in the background). For that, you'll need this dependency:
```
implementation "androidx.lifecycle:lifecycle-runtime-compose:2.6.0"
```

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginScreen.kt
```diff
@@ -13,33 +12,48 @@ import androidx.compose.ui.text.withStyle
...
+
+    LoginScreenContent(
+        state = state,
+        events = viewModel::onTriggerEvent
+    )
+
+    ManageUIEvents(
+        uiEvent = viewModel.uiEvent,
+        showDialog = {
+            state.scaffoldState.showDialog(it)
```
philipplackner, line 41: I also wonder where this `showDialog()` function comes from. From your PR it seems to be a custom scaffold state, but why did you put that in a scaffold in particular? A dialog is quite independent of a scaffold. Also curious how you manage the state of the dialog

philipplackner (approved):

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginScreen.kt
ArtemisSoftware: Will update

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/login/LoginScreen.kt
ArtemisSoftware: Because I did a custom scaffold it already has a dialog composable there as part of the TaskyScaffold.
I can see that it may be confusing if I want a dialog and I am not using the TaskyScaffold.
I will update the name of it (on a future PR) to something clearer and more generic

PR #31, opened 2023-03-20
Feature/tasks

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/domain/repositories/TaskRepository.kt
```diff
@@ -0,0 +1,13 @@
+package com.artemissoftware.tasky.agenda.domain.repositories
+
+import com.artemissoftware.core.domain.models.api.ApiNetworkResponse
+import com.artemissoftware.tasky.agenda.domain.models.AgendaItem
+
+interface TaskRepository {
+
+    suspend fun getTask(id: String): AgendaItem.Task?
+
+    suspend fun saveTaskAndSync(task: AgendaItem.Task): ApiNetworkResponse<Unit>
```
MKiperszmid, line 10: I'm not a fan of the response name "ApiNetworkResponse"

Since now domain knows it has something to do with ApiNetwork. And domain shouldn't know about that sort of stuff

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/data/remote/source/AgendaApiSource.kt
```diff
@@ -1,16 +1,20 @@
...
-}
\ No newline at end of file
+    suspend fun deleteReminder(reminderId: String): ResponseBody
+
+    suspend fun createTask(task: TaskDto): ResponseBody
+
+    suspend fun updateTask(task: TaskDto): ResponseBody
+
+    suspend fun deleteTask(taskId: String): ResponseBody
+}
```
philipplackner: This abstraction feels a bit unnecessary to me, why not just pass the Retrofit interface to your repo's implementation?
In the end the job of the repo is to combine multiple data sources and decide what to return

philipplackner (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/domain/repositories/TaskRepository.kt
ArtemisSoftware: Will make the change to DataResponse

app/src/main/java/com/artemissoftware/tasky/agenda/data/remote/source/AgendaApiSource.kt
ArtemisSoftware: AgendaApiSourceImpl.kt class handles any errors on the api calls.
When there is an error it will throw an exception that will be caught by the repository

I create this interface so the repository stays a little cleaner.

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/TaskRepositoryImpl.kt
```diff
@@ -0,0 +1,56 @@
...
+        taskDao.upsertSyncStateAndTask(task.toEntity(), TaskSyncEntity(id = task.id, syncType = syncType))
+
+        return try {
+            when (syncType) {
+                SyncType.CREATE -> {
+                    agendaApiSource.createTask(task.toDto())
+                }
+                SyncType.UPDATE -> {
+                    agendaApiSource.updateTask(task.toDto())
+                }
```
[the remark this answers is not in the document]
MKiperszmid: Right
But if it throws an exception then it never saves to the room db, right?

Which beats the purpose of being offline-first


I recommend here to save it to db, and based on whether it was synced or not, update the "syncType"

app/src/main/java/com/artemissoftware/tasky/agenda/data/remote/source/AgendaApiSource.kt
```diff
@@ -1,16 +1,20 @@
 package com.artemissoftware.tasky.agenda.data.remote.source
 
 import com.artemissoftware.tasky.agenda.data.remote.dto.ReminderDto
-import com.artemissoftware.tasky.authentication.data.remote.dto.RegistrationBodyDto
+import com.artemissoftware.tasky.agenda.data.remote.dto.TaskDto
 import okhttp3.ResponseBody
-import retrofit2.http.Body
-import retrofit2.http.Query
 
 interface AgendaApiSource {
```
MKiperszmid, line 7: What's the use of having this as an interface?

Will you implement it anywhere else?

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/TaskRepositoryImpl.kt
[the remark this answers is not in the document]
ArtemisSoftware: It is saved on line 26

app/src/main/java/com/artemissoftware/tasky/agenda/data/remote/source/AgendaApiSource.kt
ArtemisSoftware: I alreday started to implement it on this file AgendaApiSourceImpl

MKiperszmid (approved):

PR #36, opened 2023-03-21
Feature/splash view model

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/authentication/domain/usecases/AuthenticateUseCase.kt
```diff
@@ -2,19 +2,18 @@ package com.artemissoftware.tasky.authentication.domain.usecases
...
+    private val authenticationRepository: AuthenticationRepository,
+) {
 
     suspend operator fun invoke(): Resource<Unit> {
-
         val result = authenticationRepository.authenticate()
 
-        return when(result){
+        return when (result) {
             is DataResponse.Error -> {
```
philipplackner, line 17: Have you considered setting up a linting tool like ktlint which enforces a consistent formatting after building/committing, so that such changed lines wouldn't appear on the PR anymore? I've seen huge PRs contain barely anything just because there were some `Ctrl+Alt+L` presses which changed the formatting 😄

app/src/main/java/com/artemissoftware/tasky/authentication/presentation/splash/SplashScreen.kt
```diff
@@ -3,41 +3,56 @@ package com.artemissoftware.tasky.authentication.presentation.splash
...
+    SplashScreenContent(state = state)
+
+    ManageUIEvents(
+        uiEvent = viewModel.uiEvent,
+        showDialog = {
+            state.scaffoldState.showDialog(it)
+        },
+        onNavigate = {},
+    )
+}
```
philipplackner, line 32: I'd really avoid implementing a separate splash screen screen, since Android already shows a splash screen by default on later API levels. This might result in 2 splash screens being shown. Are you aware there is an official splash screen API which also allows you to do/check something while it's shown?
I show it in this video: https://www.youtube.com/watch?v=Loo4i5IrZ4Y

app/src/main/java/com/artemissoftware/tasky/authentication/domain/usecases/AuthenticateUseCase.kt
ArtemisSoftware: Since you mentioned it last week I have been using a plugin to enforce formatting.
I am doing the reformating class by class.

app/src/main/java/com/artemissoftware/tasky/MainViewModel.kt
```diff
@@ -0,0 +1,77 @@
...
+    private val _state = MutableStateFlow(MainState())
+    val state: StateFlow<MainState> = _state
+
+    init {
+        authenticate()
+    }
+
+    private fun authenticate() {
+        viewModelScope.launch {
+            delay(1500)
```
MKiperszmid, line 37: What is this delay for?

app/src/main/java/com/artemissoftware/tasky/MainState.kt
```diff
@@ -0,0 +1,10 @@
+package com.artemissoftware.tasky
+
+import com.artemissoftware.core.presentation.composables.scaffold.TaskyScaffoldState
+import com.artemissoftware.tasky.navigation.Destination
+
+data class MainState(
+    val screenToNavigate: Destination? = null,
```
MKiperszmid, line 7: I personally dont like having the viewmodel decide which screen to go


I like exposing something like a Boolean, and then decide which one to show when displaying the NavHost

app/src/main/java/com/artemissoftware/tasky/MainState.kt
```diff
@@ -0,0 +1,10 @@
...
+    val isLoading: Boolean = true,
+    val scaffoldState: TaskyScaffoldState = TaskyScaffoldState()
```
MKiperszmid, line 9: Whats the use of this state?

Cant it be shown only on the screen without the viewmodel knowing about it?

Im not seeing it being used on the viewmodel

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/MainState.kt
ArtemisSoftware: This state is used in the following way

 isLoading -> controlls if the splash screen should still be visible (I will change the name to showSplash)
 
 scaffoldState -> is used to display error dialogs. An example is if your API Key is not valid it will display a dialog with a message
 
 screenToNavigate ->This one decides to wich screen it should navigate to

app/src/main/java/com/artemissoftware/tasky/MainViewModel.kt
ArtemisSoftware: Its just so the user sees the splash screen for a while.
If the authenticate api call is too fast the user at least enjoys the splash screen for a while.

I changed its place

app/src/main/java/com/artemissoftware/tasky/MainState.kt
ArtemisSoftware: But how do you deal with a situation that after the splash you can go to 5 different screens?

I think the viewmodel should decide about the navigation since it is already controlling a lot of things regarding the screen.
If the viewmodel already has responsabilities why not delegate this one to it and make the screen unaware of those things?

app/src/main/java/com/artemissoftware/tasky/MainViewModel.kt
```diff
@@ -0,0 +1,74 @@
...
+                    result.exception?.let { exception ->
+                        when (exception) {
+                            AuthenticationException.UserNotAuthenticated -> {
+                                _state.update {
+                                    it.copy(showSplash = false, destinationAfterSplash = Destination.Login)
+                                }
+                            }
+                            else -> {
+                                _state.update { it.copy(showSplash = false) }
+                                sendUiEvent(UiEvent.ShowDialog(getDialogData(ex = exception, reloadEvent = { authenticate() })))
```
MKiperszmid, line 52: What if it returns an exception other than UserNotAuthenticated?
It will call these again

And if it returns a different exception?
It will call it again

And again
And again

So you have an infinite loop

You can probably try adding a retry limit, so if it has the same error 3 times, then redirect them to Login

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/MainViewModel.kt
philipplackner: Is the `authenticateRetries` considered here for `reloadEvent` or only for updating the state?

philipplackner (approved):

app/src/main/java/com/artemissoftware/tasky/MainViewModel.kt
ArtemisSoftware: It is for the reloadEvent

app/src/main/java/com/artemissoftware/tasky/MainState.kt
philipplackner: The ViewModel will very often determine the screen to navigate, but usually rather indirectly (by updating a state `isLoggedIn` for example to get to the logged in screen). Don't see a big downside of this, just that you're not so flexible with popping the backstack while navigating (unless `Destination` supports that)

PR #37, opened 2023-03-22
Feature/agenda options

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/presentation/dashboard/composables/AgendaTopBar.kt
```diff
@@ -91,3 +87,6 @@ private fun AgendaTopBarPreview() {
         )
     }
 }
+
+
+
```
MKiperszmid, line 92: This can be removed

core/src/main/java/com/artemissoftware/core/presentation/composables/menu/TaskyPopUpMenu.kt
```diff
@@ -0,0 +1,117 @@
...
+                        expanded = false
+                        onOptionSelected.invoke(item)
+                    },
+                ) {
+                    menuOption(item)
+                }
+
+                if (addDivider && !isLast) {
+                    Divider(color = Light)
+                }
```
MKiperszmid, line 80: I'm pretty sure this doesn't need to be added since DropdownMenu already handles it (I might be mistaken)

MKiperszmid (approved):

core/src/main/java/com/artemissoftware/core/presentation/composables/menu/TaskyPopUpMenu.kt
ArtemisSoftware: I tested it, the dropdown menu does not add dividers.
They have to be added manually

PR #38, opened 2023-03-22
Feature/di authentication

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/authentication/di/NetworkModule.kt
```diff
@@ -0,0 +1,35 @@
...
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+import okhttp3.OkHttpClient
+import retrofit2.Retrofit
+import retrofit2.converter.moshi.MoshiConverterFactory
+import javax.inject.Singleton
+
+@Module
+@InstallIn(SingletonComponent::class)
+object NetworkModule {
```
MKiperszmid, line 17: Personal opinion:

To avoid having tons of modules, I usually have 1 "Data Module", since usually all the data stuff is used at the same time

In this case, the AuthAPI and AuthRepository

core/src/main/java/com/artemissoftware/core/di/TaskyNetworkModule.kt
```diff
@@ -0,0 +1,34 @@
...
+import java.util.concurrent.TimeUnit
+import javax.inject.Singleton
+
+@Module
+@InstallIn(SingletonComponent::class)
+object TaskyNetworkModule {
+
+    @Provides
+    @Singleton
+    fun provideHttpClient(getUserUseCase: GetUserUseCase): OkHttpClient {
```
MKiperszmid, line 22: Another personal opinion:

Here I prefer injecting a SharedPreference, and get the user in the interceptor

Usually you won't use UseCases on the Data layer

MKiperszmid (approved):

PR #39, opened 2023-03-24
Feature/event

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/data/mappers/EventMapper.kt
```diff
@@ -0,0 +1,125 @@
...
+
+fun AgendaItem.Event.toEntity(): EventEntity {
+    return EventEntity(
+        title = title,
+        description = this.description ?: "",
+        id = this.id,
+        remindAt = remindAt.toLong(),
+        to = to.toLong(),
+        from = from.toLong(),
+        isUserEventCreator = true, // TODO: not sure about it
```
MKiperszmid, line 70: I personally just saved the host id, and then verify if its the host or not based on if the id is the same as the current's user id

Mostly to save 1 less parameter in the table
But saving it like you did should work just fine

app/src/main/java/com/artemissoftware/tasky/agenda/data/remote/source/AgendaApiSource.kt
```diff
@@ -43,4 +48,48 @@ class AgendaApiSource constructor(private val taskyAgendaApi: TaskyAgendaApi) {
...
+        val files = imageFiles.mapIndexed { index, file ->
+            MultipartBody.Part
+                .createFormData(
+                    name = "photo_$index",
+                    filename = file.name,
+                    body = file.asRequestBody(),
+                )
+        }
+
+        return HandleApi.safeApiCall {
```
MKiperszmid, line 66: Since you are uploading images, this can take a while

Maybe it would be better to use a WorkManager rather than just coroutines

app/src/main/java/com/artemissoftware/tasky/agenda/data/remote/source/AgendaApiSource.kt
```diff
@@ -43,4 +48,48 @@ class AgendaApiSource constructor(private val taskyAgendaApi: TaskyAgendaApi) {
...
```
MKiperszmid, line 85: Same here about the WorkManager

app/src/main/java/com/artemissoftware/tasky/agenda/data/remote/source/AgendaApiSource.kt
```diff
@@ -43,4 +48,48 @@ class AgendaApiSource constructor(private val taskyAgendaApi: TaskyAgendaApi) {
...
+            taskyAgendaApi.updateEvent(body = event, files = files)
+        }
+    }
+
+    suspend fun deleteEvent(eventId: String): ResponseBody {
+        return HandleApi.safeApiCall {
+            taskyAgendaApi.deleteEvent(eventId)
```
MKiperszmid, line 92: This one is correct to use coroutines
Since you wont be uploading images

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/EventRepositoryImpl.kt
```diff
@@ -0,0 +1,61 @@
...
+    override suspend fun saveEventAndSync(event: AgendaItem.Event): DataResponse<Unit> {
+        val syncType = if (event.syncState == SyncType.SYNCED) SyncType.UPDATE else event.syncState
+        eventDao.upsertSyncStateAndEvent(
+            eventEntity = event.toEntity(),
+            photoEntities = event.toPhotoEntities(),
+            eventAttendeeEntities = event.toAttendeeEntities(),
+            eventSyncEntity = EventSyncEntity(id = event.id, syncType = syncType),
+        )
+
+        val imageFiles = event.photos.filter { it.local != null }.map { it.local!! }
```
MKiperszmid, line 31: I think you could just:

.mapNotNull {it.local}

(Or filterNotNull, cant remember correctly)

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/AgendaItem.kt
```diff
@@ -45,4 +46,27 @@ sealed class AgendaItem(
...
+        var description: String? = null,
+        var remindAt: LocalDateTime,
+        var to: LocalDateTime,
+        var from: LocalDateTime,
+        var host: String,
+        var isDone: Boolean = false,
+        var photos: List<Photo>,
+        var deletedPhotosIds: List<String> = emptyList(),
+        var attendees: List<Attendee>,
+        var isGoing: Boolean, // TODO: where do I get this value? need to check
```
MKiperszmid, line 62: I cant remember the api docs exactly

But arent you receiving something like the list of attendees with the isGoing true/false?

If so, you could check if your id/email is in the list of going or not going

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/Photo.kt
```diff
@@ -1,18 +1,19 @@
 package com.artemissoftware.tasky.agenda.domain.models
 
+import java.io.File
+
 data class Photo(
     val key: String,
-    val url: String
-    //val localImages: String // TODO: must support this too
-){
+    val url: String,
```
MKiperszmid, line 7: As far as I remember, Photos will have either a url or a local uri

Here you could do something like what you did with AgendaItem
A sealed class that has 2 children. RemotePhoto and LocalPhoto

MKiperszmid (approved):

ArtemisSoftware:
Out of date

PR #41, opened 2023-03-27
Feature/agenda options

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/composables/DatePickerDialog.kt
```diff
@@ -0,0 +1,28 @@
...
+import java.time.LocalDate
+
+object DateTimePicker {
+
+    fun datePickerDialog(
+        context: Context,
+        date: MutableState<LocalDate>,
+        onDateSelected: (LocalDate) -> Unit = {},
+    ): DatePickerDialog {
+        return DatePickerDialog(
```
MKiperszmid, line 16: Why do you have an object class that does this logic?
Can't you just call it in the screen directly?

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/dashboard/AgendaScreen.kt
```diff
@@ -51,6 +57,10 @@ fun AgendaScreen(
     state: AgendaState,
     events: (AgendaEvents) -> Unit,
 ) {
+    val context = LocalContext.current
+
+    val date = remember { mutableStateOf(LocalDate.now()) }
```
MKiperszmid, line 62: Why not have it on the state?

It makes it more consistent to save all the state from a screen in one place
In this case, on the viewmodel

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/composables/DatePickerDialog.kt
MKiperszmid: Yea but can't it be called directly on the Detail Screen?

PR #44, opened 2023-03-28
Feature/di agenda

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/data/remote/source/AgendaApiSource.kt
```diff
@@ -5,8 +5,9 @@ import com.artemissoftware.tasky.agenda.data.remote.TaskyAgendaApi
 import com.artemissoftware.tasky.agenda.data.remote.dto.ReminderDto
 import com.artemissoftware.tasky.agenda.data.remote.dto.TaskDto
 import okhttp3.ResponseBody
+import javax.inject.Inject
 
-class AgendaApiSource constructor(private val taskyAgendaApi: TaskyAgendaApi) {
+class AgendaApiSource @Inject constructor(private val taskyAgendaApi: TaskyAgendaApi) {
```
MKiperszmid, line 10: You are creating the AgendaApiSource  on NetworkModule, so you don't need @Inject constructor

You can just:

```
class AgendaApiSource(private val taskyAgendaApi: TaskyAgendaApi)
```

philipplackner (approved):

core/src/main/java/com/artemissoftware/core/util/Constants.kt
```diff
@@ -0,0 +1,6 @@
+package com.artemissoftware.core.util
+
+object Constants {
+
+    const val TASKY_DATABASE = "tasky_db"
```
MKiperszmid, line 5: You are only using this constant once. So no need for a constant

MKiperszmid (approved):

PR #48, opened 2023-03-31
TaskDetailScreen

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/taskdetail/TaskDetailScreen.kt
```diff
@@ -0,0 +1,208 @@
...
+                    events(DetailEvents.PopBackStack)
+                },
+                backGroundColor = Black,
+                allCaps = true,
+                title = String.format(
+                    resources.getString(R.string.edit_title_with_argument),
+                    resources.getString(
+                        R.string.task,
+                    ),
+                ),
```
philipplackner, line 62: Instead of `resources` you can just use the built in composables: `stringResource(R.string.your_id, arg1, arg2)`

ArtemisSoftware: Refactored

PR #50, opened 2023-04-03
Feature/remainder view model part 2

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/reminderdetail/ReminderDetailScreen.kt
```diff
@@ -0,0 +1,254 @@
...
+            navigator.popBackStack()
+        },
+    )
+}
+
+@Composable
+fun ReminderDetailScreenContent(
+    state: DetailState,
+    events: (DetailEvents) -> Unit,
+    notifications: List<Notification>,
```
MKiperszmid, line 94: Remember to pass a Modifier

MKiperszmid: Also, what's this List\<Notification\> ?

Wouldn't this be in the db? (Which can be obtained when getting the reminder by id from the db)

core/src/main/java/com/artemissoftware/core/data/database/TaskyDatabase.kt
```diff
@@ -26,4 +32,14 @@ abstract class TaskyDatabase : RoomDatabase() {
...
+
+    companion object {
+        fun getInstance(context: Context) = Room.databaseBuilder(
+            context,
+            TaskyDatabase::class.java,
+            "tasky_db",
+        )
+            .addCallback(TaskyDatabaseCallback(context))
+            .build()
+    }
```
MKiperszmid, line 44: I personally create this on the Hilt Module, so I have all the creations in one place

core/src/main/java/com/artemissoftware/core/data/database/dao/ReminderDao.kt
```diff
@@ -4,22 +4,19 @@ import androidx.room.*
 import com.artemissoftware.core.data.database.entities.ReminderEntity
 import com.artemissoftware.core.data.database.entities.ReminderSyncEntity
 import com.artemissoftware.core.data.database.entities.relations.ReminderAndSyncState
-import com.artemissoftware.core.domain.SyncType
 
 @Dao
 interface ReminderDao {
-
+    @Transaction
```
MKiperszmid, line 10: Usually @Transaction is used when you are doing more than 1 query at a time (Like a JOIN or similar), so I'm pretty sure it's not necessary here

MKiperszmid: Tho im not understanding something

upsert will insert a ReminderEntity
This getter will return a ReminderAndSyncState

But you are never making a join or something that get's the ReminderEntity + SyncState and combining them both

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/reminderdetail/ReminderDetailScreen.kt
```diff
@@ -0,0 +1,254 @@
...
+import com.ramcosta.composedestinations.navigation.DestinationsNavigator
+import com.ramcosta.composedestinations.result.ResultRecipient
+import com.ramcosta.composedestinations.result.getOr
+
+@Destination
+@Composable
+fun ReminderDetailScreen(
+    navigator: DestinationsNavigator,
+    viewModel: ReminderDetailViewModel = hiltViewModel(),
+    reminderId: String? = null,
```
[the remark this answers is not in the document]
ArtemisSoftware: Compose Destinations allows me to get parameters passed from one screen to the other  on the viewmodel using savedStateHandle.

reminderId can be null when on the agenda screen you say you want to create a new reminder

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/reminderdetail/ReminderDetailScreen.kt
ArtemisSoftware: This content Composables dont need to have a modifier. They are used so I can have a working preview on the android studio of a particular screen.
If you notice ReminderDetailScreen has the viewmodel as a parameter and this breaks the preview. By creating a ReminderDetailScreenContent composable I can circunvent this issue. I forgot to make it private


List<Notification> is the list of notifications availabe to shown to the user when he presses the notification.
I created a specific usecase (GetNotificationsUsecase) the fetches them from the db. Its is being used by the view model

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/reminderdetail/ReminderDetailScreen.kt
```diff
@@ -0,0 +1,254 @@
...
+                            .padding(horizontal = 16.dp),
+                    ) {
+                        Column(
+                            modifier = Modifier
+                                .align(Alignment.TopCenter),
+                            horizontalAlignment = Alignment.CenterHorizontally,
+                        ) {
+                            AssignmentHeader(
+                                agendaItemType = AgendaItemType.Reminder(),
+                                title = state.agendaItem?.itemTitle ?: "",
```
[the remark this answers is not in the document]
ArtemisSoftware: I am already changing this. On the next PR this issue as been addressed

core/src/main/java/com/artemissoftware/core/data/database/TaskyDatabase.kt
ArtemisSoftware: For what I tried to achieve I was not able to have it on the DatabaseMoule.

I want to prepopulate the database with all the notifications. But all the examples I found of prepopulating the database refered to something like this.

The main issue is that the TaskyDatabaseCallback which will prepopulate the database needs an instance of the database.

Do you know of another way of doing this?

core/src/main/java/com/artemissoftware/core/data/database/TaskyDatabaseCallback.kt
```diff
@@ -0,0 +1,49 @@
...
+
+import android.content.Context
+import androidx.room.RoomDatabase
+import androidx.sqlite.db.SupportSQLiteDatabase
+import com.artemissoftware.core.data.database.entities.NotificationWarningEntity
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.launch
+
+class TaskyDatabaseCallback(private val context: Context) : RoomDatabase.Callback() {
```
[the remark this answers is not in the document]
ArtemisSoftware: No,
the TaskyDatabaseCallback will prepopulate the database.
See above comment

core/src/main/java/com/artemissoftware/core/data/database/dao/ReminderDao.kt
ArtemisSoftware: I added the  @transaction because the compiler was giving me a warning

Here is the warning

core\data\database\dao\ReminderDao.java:11: warning: The return value includes a POJO with a @Relation. It is usually desired to annotate this method with @Transaction to avoid possibility of inconsistent results between the POJO and its relations. See https://developer.android.com/reference/androidx/room/Transaction.html for details.
    public abstract com.artemissoftware.core.data.database.entities.relations.ReminderAndSyncState getReminderAndSyncState(@org.jetbrains.annotations.NotNull()
    
    
I do combine   ReminderEntity + SyncStat on the ReminderAndSyncState class. There I create relations on between both

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/reminderdetail/ReminderDetailScreen.kt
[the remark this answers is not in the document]
MKiperszmid: Gotcha, sounds good

But why are you passing the id here, instead of using savedStateHandle directly on the viewmodel?

core/src/main/java/com/artemissoftware/core/data/database/TaskyDatabase.kt
MKiperszmid: Mmm never had to use a callback, so never tried it

But im curious
Why do you need a callback? I know you prepopulate the db, but usually when a user logs in, it shouldnt have random info. Rather just the info from their api

core/src/main/java/com/artemissoftware/core/data/database/TaskyDatabaseCallback.kt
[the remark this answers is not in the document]
MKiperszmid: Replied above

core/src/main/java/com/artemissoftware/core/data/database/dao/ReminderDao.kt
MKiperszmid: Gotcha
Yeah that one uses a join in the back

Sounds good

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/reminderdetail/ReminderDetailScreen.kt
ArtemisSoftware: I did not explain correctly.

The viewmodel has the GetNotificationsUsecase on its constructor. When I get the notifications from the database I put it on a mutablestate. I still need to pass it as parameter so I can populate the composable with all the options

core/src/main/java/com/artemissoftware/core/data/database/TaskyDatabase.kt
ArtemisSoftware: From what I understood the callback will be "called" on the onCreate of the database

@philipplackner could you give some feedback on this issues?

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/reminderdetail/ReminderDetailScreen.kt
MKiperszmid: Right. But cant that information be added to the DetailState?

app/src/main/java/com/artemissoftware/tasky/agenda/composables/assignment/AssignmentNotification.kt
```diff
@@ -1,57 +1,58 @@
...
 import com.artemissoftware.tasky.util.VisibilityTransitions
 
 @Composable
 fun AssignmentNotification(
-    selectedNotification: Notification,
-    onNotificationSelected: (Notification) -> Unit,
-    notificationOptions: List<Notification>,
+    selectedNotification: NotificationType,
+    onNotificationSelected: (NotificationType) -> Unit,
+    notificationOptions: List<NotificationType>,
```
MKiperszmid, line 30: Do you really need pass this variable?
Since you will always use the same list (NotificationType.values()) right?

core/src/main/java/com/artemissoftware/core/domain/models/agenda/NotificationType.kt
```diff
@@ -0,0 +1,53 @@
...
+    ),
+    ONE_DAY_BEFORE(
+        minutesBefore = 24 * 60,
+        description = UiText.StringResource(R.string.one_day_before),
+    ),
+    ;
+
+    companion object {
+
+        fun defaultNotification() = values().find { it.isDefault } ?: THIRTY_MINUTES_BEFORE
```
MKiperszmid, line 36: You could just return a default value, no need for all these logic of storing a default attribute and such

MKiperszmid (approved):

core/src/main/java/com/artemissoftware/core/domain/models/agenda/NotificationType.kt
ArtemisSoftware: Sure

PR #51, opened 2023-04-05
Feature/task view model

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/taskdetail/TaskDetailViewModel.kt
```diff
@@ -0,0 +1,188 @@
...
+                    it.copy(
+                        agendaItem = AgendaItem.Task(),
+                    )
+                }
+            }
+        }
+    }
+
+    private fun saveTask() {
+        (_state.value.agendaItem as? AgendaItem.Task)?.let { item ->
```
MKiperszmid, line 172: I'm not a fan of this cast, since you should always have a Task here, and nothing else

One way is to have the Item be explicitly mentioned as a Task on the state

That's why I recommend having one state for each screen, instead of sharing one state on multiple screens

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/taskdetail/TaskDetailViewModel.kt
```diff
@@ -0,0 +1,188 @@
...
+
+            with(_state.value) {
+                item.itemTitle = title
+                item.itemDescription = description
+                item.itemRemindAt = NotificationType.remindAt(time = startDate, notificationType = notification)
+                item.starDate = startDate
+                item.isDone = (specification as DetailSpecification.Task).isDone
```
MKiperszmid, line 179: Same here (comment above)

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/taskdetail/TaskDetailViewModel.kt
ArtemisSoftware: Refactored

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/AgendaItem.kt
```diff
@@ -10,7 +10,7 @@ sealed class AgendaItem(
     var itemTitle: String,
     var itemDescription: String? = null,
     var itemRemindAt: LocalDateTime,
-    val starDate: LocalDateTime,
+    var starDate: LocalDateTime,
```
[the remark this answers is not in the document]
ArtemisSoftware: Refactored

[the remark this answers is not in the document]
ArtemisSoftware: The ones that are var are the ones that can change while editing an agendaItem

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/reminderdetail/ReminderDetailState.kt
```diff
@@ -0,0 +1,15 @@
+package com.artemissoftware.tasky.agenda.presentation.detail.reminderdetail
+
+import com.artemissoftware.core.domain.models.agenda.NotificationType
+import com.artemissoftware.tasky.agenda.domain.models.AgendaItem
+import java.time.LocalDateTime
+
+data class ReminderDetailState(
+    val isLoading: Boolean = false,
+    val isEditing: Boolean = false,
+    val agendaItem: AgendaItem.Reminder? = null,
```
MKiperszmid, line 10: When it's null, you set a default value: https://github.com/ArtemisSoftware/Tasky/pull/51/files#diff-ddde38337076cab4ebc9ee543c969af2ce248e1aa88cefa335c505c91b1fe8e5R151

You can just do it here instead, and avoid having nullable ;)

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/taskdetail/TaskDetailState.kt
```diff
@@ -0,0 +1,16 @@
+package com.artemissoftware.tasky.agenda.presentation.detail.taskdetail
+
+import com.artemissoftware.core.domain.models.agenda.NotificationType
+import com.artemissoftware.tasky.agenda.domain.models.AgendaItem
+import java.time.LocalDateTime
+
+data class TaskDetailState(
+    val isLoading: Boolean = false,
+    val isEditing: Boolean = false,
+    val agendaItem: AgendaItem.Task? = null,
```
MKiperszmid, line 10: Same as what I mentioned on ReminderDetailState

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/reminderdetail/ReminderDetailState.kt
ArtemisSoftware: Good call

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/taskdetail/TaskDetailViewModel.kt
```diff
@@ -0,0 +1,183 @@
...
+            it.copy(
+                isEditing = !_state.value.isEditing,
+            )
+        }
+    }
+
+    private fun toggleIsDone() {
+        _state.update {
+            it.copy(
+                isDone = !_state.value.isDone,
```
philipplackner, line 66: `_state.update` gives you a reference to the current state, so you can do `isDone = !it.isDone`

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/taskdetail/TaskDetailViewModel.kt
```diff
@@ -0,0 +1,183 @@
...
+                    }
+                }
+            } ?: run {
+                _state.update {
+                    it.copy(
+                        agendaItem = AgendaItem.Task(),
+                    )
+                }
+            }
+        }
```
philipplackner, line 163: Here you could use an early return to get rid of one indentation:
```
val id = savedStateHandle.get<String>("taskId")
if(id == null) {
    _state.update {
        it.copy(
            agendaItem = AgendaItem.Task(),
        )
    }
    return
}

// ID is certainly non-null from this point
```

philipplackner (approved):

PR #53, opened 2023-04-06
Feature/alarm part 2

ArtemisSoftware:


core/src/main/java/com/artemissoftware/core/util/TaskyNotification.kt
```diff
@@ -0,0 +1,40 @@
...
+import android.app.NotificationManager
+import android.app.PendingIntent
+import android.content.Context
+import androidx.core.app.NotificationCompat
+import com.artemissoftware.core.R
+
+object TaskyNotification {
+
+    fun sendNotification(context: Context, id: String, title: String, body: String, pendingIntent: PendingIntent) {
+        createChannel(context = context)
```
MKiperszmid, line 13: Are you targeting api 24 and up?
If you are targeting api's lower than 24, then the channel part will crash the app since that only works on API 24+

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/taskdetail/TaskDetailViewModel.kt
```diff
@@ -154,6 +155,9 @@ class TaskDetailViewModel @Inject constructor(
         agendaItem.remindAt = NotificationType.remindAt(time = startDate, notificationType = notification)
         agendaItem.time = startDate
         agendaItem.isDone = isDone
+        savedStateHandle.get<String>("taskId")?.let {
```
MKiperszmid, line 158: Here you could use a constant for "taskId"

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/AgendaItem.kt
```diff
@@ -41,7 +41,7 @@ sealed class AgendaItem(
         var isDone: Boolean = false,
         var remindAt: LocalDateTime = NotificationType.remindAt(time, NotificationType.defaultNotification()),
         var notification: NotificationType = NotificationType.getNotification(remindAt = remindAt, startDate = time),
-        val syncState: SyncType = SyncType.CREATE,
+        var syncState: SyncType = SyncType.CREATE,
```
MKiperszmid, line 44: I'm not a big fan of having data classes with var. Mostly because it keeps the immutability. One of Kotlin's benefits

Why do you need a var in this case?

MKiperszmid (approved):

core/src/main/java/com/artemissoftware/core/util/TaskyNotification.kt
ArtemisSoftware: I am targeting api 28.

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/taskdetail/TaskDetailViewModel.kt
ArtemisSoftware: Refactored

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/AgendaItem.kt
ArtemisSoftware: Refactored

app/src/main/java/com/artemissoftware/tasky/agenda/data/alarm/AlarmReceiver.kt
```diff
@@ -1,11 +1,52 @@
...
+                    body = body,
+                    pendingIntent = getPendingIntent(context = currentContext),
+                )
+            }
+        }
+    }
+
+    private fun getPendingIntent(context: Context): PendingIntent {
+        // TODO: how to get the destinations here if we are on the data module that does not know about navigation?
+        val deeplink = "https://tasky.com".toUri()
```
philipplackner, line 34: I'd say it's fine to use the destination here like this: `"tasky://event_detail?eventId=${item.id}".toUri()`
It would also be fine IMO to put this receiver in presentation if it's directly presentation related (which showing a notification is in my opinion)

philipplackner (approved):

PR #55, opened 2023-04-10
Feature/attendee api call

ArtemisSoftware:


core/src/main/java/com/artemissoftware/core/domain/AgendaException.kt
```diff
@@ -0,0 +1,7 @@
+package com.artemissoftware.core.domain
+
+sealed class AgendaException : ValidationException() {
+
+    object AttendeeError : AgendaException()
```
MKiperszmid, line 5: Will this exception ever be thrown?
If so, what will be the use of it?

Whats the difference between a regular exception?

core/src/main/java/com/artemissoftware/core/presentation/mappers/ValidationExceptionMapper.kt
```diff
@@ -20,6 +20,11 @@ fun ValidationException.toUiText(): UiText {
         is ValidationException.DataError -> {
             this.uiText
         }
+        AgendaException.AttendeeDoesNotExist -> {
+            UiText.StringResource(R.string.attendee_not_exist)
+        }
+        AgendaException.AttendeeError -> {
+            UiText.StringResource(R.string.error_occurred_during_attendee_search)
```
MKiperszmid, line 27: Here if you remove AttendeeError (previous comment), you could simply have an if statement

If error is AttendeeDoesNotExist, do something. Else, do other thing

MKiperszmid (approved):

core/src/main/java/com/artemissoftware/core/domain/AgendaException.kt
ArtemisSoftware: This exception will be thrown when the GetAttendeeUseCase has an error

PR #56, opened 2023-04-10
Feature/alarm boot

ArtemisSoftware:


core/src/main/java/com/artemissoftware/core/data/database/dao/ReminderDao.kt
```diff
@@ -11,6 +11,10 @@ interface ReminderDao {
     @Query("SELECT * FROM reminderEntity WHERE id = :id")
     fun getReminderAndSyncState(id: String): ReminderAndSyncState?
 
+    @Transaction
+    @Query("SELECT * FROM reminderEntity WHERE time >= :currentDate")
+    suspend fun getRemindersToSetAlarm(currentDate: Long): List<ReminderAndSyncState>
```
MKiperszmid, line 16: The parameter is "currentDate". But wouldnt it make more sense to name it currentTime?
Since you arent checking the date itself

MKiperszmid (approved):

core/src/main/java/com/artemissoftware/core/data/database/dao/ReminderDao.kt
ArtemisSoftware: Refactored

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/AgendaRepositoryImpl.kt
```diff
@@ -1,5 +1,22 @@
...
+import com.artemissoftware.tasky.agenda.data.mappers.toAgendaItem
+import com.artemissoftware.tasky.agenda.domain.models.AgendaItem
 import com.artemissoftware.tasky.agenda.domain.repositories.AgendaRepository
 
-class AgendaRepositoryImpl() : AgendaRepository
+class AgendaRepositoryImpl(
+    private val reminderDao: ReminderDao,
+    private val taskDao: TaskDao,
+) : AgendaRepository {
+    override suspend fun getAgendaToUpdateAlarms(currentDate: Long): List<AgendaItem> {
```
philipplackner, line 13: I'd keep this independent of the use case ("to update alarms"), since it might be used for more than that

app/src/main/java/com/artemissoftware/tasky/agenda/domain/usecase/agenda/GetAgendaToUpdateAlarmsUseCase.kt
```diff
@@ -0,0 +1,16 @@
+package com.artemissoftware.tasky.agenda.domain.usecase.agenda
+
+import com.artemissoftware.tasky.agenda.domain.models.AgendaItem
+import com.artemissoftware.tasky.agenda.domain.repositories.AgendaRepository
+import java.time.LocalDate
+import java.time.ZoneId
+import javax.inject.Inject
+
+class GetAgendaToUpdateAlarmsUseCase @Inject constructor(
```
philipplackner, line 9: Same here, I think just `GetFutureAgendaItems` or so would be enough

PR #57, opened 2023-04-11
Feature/agenda reminder logic

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/ReminderRepositoryImpl.kt
```diff
@@ -21,6 +25,15 @@ class ReminderRepositoryImpl constructor(
         return reminderDao.getReminderAndSyncState(id)?.toAgendaItem()
     }
 
+    override fun getReminders(date: LocalDate): Flow<List<AgendaItem.Reminder>> {
+        val initialDate = date.atStartOfDay().atZone(ZoneId.systemDefault()).toInstant().toEpochMilli()
+        val endDate = date.atStartOfDay().plusDays(1).atZone(ZoneId.systemDefault()).toInstant().toEpochMilli()
```
MKiperszmid, line 30: Maybe this ".atZone(ZoneId.systemDefault()).toInstant().toEpochMilli()" could be added on a Util class or as an Extension, since you'll use it a couple of times (for event and task also)

core/src/main/java/com/artemissoftware/core/presentation/composables/textfield/TaskyTextField.kt
```diff
@@ -34,6 +40,10 @@ fun TaskyTextField(
         ),
         onValueChange = onValueChange,
     )
+
+    LaunchedEffect(Unit) {
+        focusRequester.requestFocus()
+    }
```
MKiperszmid, line 46: What would happen if you have more than 1 TaskyTextField on a screen?
Won't this make it so they fight for the focus?

gradle.properties
```diff
@@ -21,4 +21,4 @@ kotlin.code.style=official
 # resources declared in the library itself and none from the library's dependencies,
 # thereby reducing the size of the R class for that library
 android.nonTransitiveRClass=true
-API_KEY = "thekey"
+API_KEY = "dbed1a0790fd4ad2a18efed4cfd7f67b"
```
MKiperszmid, line 24: You uploaded your API key publicly
Remember to never do this

https://www.youtube.com/watch?v=-2ckvIzs0nU

PR #63, opened 2023-04-14
Feature/event pictures compress

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/domain/usecase/event/ValidatePicturesUseCase.kt
```diff
@@ -0,0 +1,39 @@
...
+            else {
+                list.add(picture)
+            }
+        }
+
+        if (rejectedPictures != 0) {
+            emit(Resource.Error(exception = AgendaException.NotValidPictures(UiText.DynamicString("$rejectedPictures photos were skipped because they were too large"))))
+        }
+
+        emit(Resource.Success(list))
```
MKiperszmid, line 37: A better way to handle this could be to return either success or error
Not both

And on error, return the list of pictures that have the issue, so you can somehow showcase in the UI which photos are too large to be uploaded 

And if you only return one thing, you can remove using Flow

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailState.kt
```diff
@@ -2,23 +2,23 @@ package com.artemissoftware.tasky.agenda.presentation.detail.eventdetail
...
+import com.artemissoftware.tasky.agenda.domain.models.Picture
 import java.time.LocalDateTime
 
 data class EventDetailState(
     val isLoading: Boolean = false,
     val isEditing: Boolean = false,
     val eventCreatorId: String = "",
     val visitorOption: VisitorOptionType = VisitorOptionType.ALL,
-    // val agendaItem: AgendaItem.Event = AgendaItem.Event(), // TODO add when AgendaItem.Event() is ready
+    val agendaItem: AgendaItem.Event = AgendaItem.Event(),
```
MKiperszmid, line 17: Do you need this for the state tho?

EventDetailState should only work for Events. Nothing else

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailViewModel.kt
```diff
@@ -85,10 +90,35 @@ class EventDetailViewModel @Inject constructor(
...
+    }
+
+    private fun addPicture(uri: Uri) = with(_state) {
+        update {
+            val pictures = it.pictures.toMutableList()
+            pictures.add(Picture.Local(uri = uri.toString()))
+
+            it.copy(
+                pictures = pictures,
+            )
```
MKiperszmid, line 118: This can be replaced with:
```
it.copy(
    pictures = it.pictures + Picture.Local(uri = uri.toString())
)
```

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/Picture.kt
```diff
@@ -0,0 +1,12 @@
+package com.artemissoftware.tasky.agenda.domain.models
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+sealed class Picture(val source: String) : Parcelable {
+
+    data class Local(val uri: String) : Picture(source = uri)
```
philipplackner, line 9: You might also need a unique identifier for local photos because if the user adds the same photo twice and then wants to delete one, they would delete both if you identify them via Uri

app/src/main/java/com/artemissoftware/tasky/agenda/domain/usecase/event/ValidatePicturesUseCase.kt
philipplackner: To keep things simple, the requirement is simply to show how many pics were skipped. In this case, I'd also rather find it confusing to have both error and success emissions. What about creating your own result class for this which contains simply the list of added pictures and the count of rejected images?

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailState.kt
philipplackner: But isn't that what he's doing? As far as I understand he changes `agendaItem` whenever the user edits something on that screen

philipplackner (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailState.kt
MKiperszmid: You are right. I got confused thinking it was only used to define another thing

app/src/main/java/com/artemissoftware/tasky/agenda/domain/usecase/event/ValidatePicturesUseCase.kt
ArtemisSoftware: Refactored

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailViewModel.kt
ArtemisSoftware: Refactored

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/Picture.kt
ArtemisSoftware: Updated

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/Picture.kt
```diff
@@ -0,0 +1,12 @@
+package com.artemissoftware.tasky.agenda.domain.models
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+sealed class Picture(val id: String, val source: String) : Parcelable {
```
MKiperszmid, line 7: Whats id?

Source is either the uri or the url
But id?

Local only has uri
Remote has key and url

app/src/main/java/com/artemissoftware/tasky/agenda/domain/usecase/event/ValidatePicturesUseCase.kt
```diff
@@ -0,0 +1,33 @@
...
+            val imageSize = imageCompressor.compressUri(uri = picture.uri).size
+
+            if (imageSize > ImageSizeConstants.ONE_MEGA_BYTE) {
+                ++rejectedPictures
+            } else {
+                list.add(picture)
+            }
+        }
+
+        return Resource.Success(
```
MKiperszmid, line 29: You could just return a PictureValidation. No need for a Resource

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/Picture.kt
ArtemisSoftware: @philipplackner suggested I should also add an Id

app/src/main/java/com/artemissoftware/tasky/agenda/domain/usecase/event/ValidatePicturesUseCase.kt
ArtemisSoftware: Refactored

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/Picture.kt
philipplackner: @MKiperszmid If you add the same local photo twice, it'll have the same Uri. If you then want to delete one of them, they don't have unique identifiers and you'll end up deleting both

PR #66, opened 2023-04-18
Feature/event db

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/EventRepositoryImpl.kt
```diff
@@ -1,28 +1,43 @@
...
+
     override suspend fun saveEventAndSync(event: AgendaItem.Event): DataResponse<Unit> {
         val syncType = if (event.syncState == SyncType.SYNCED) SyncType.UPDATE else event.syncState
-        eventDao.upsertSyncStateAndEvent(
-            eventEntity = event.toEntity(),
-            attendees = event.attendees.map { it.toEntity(eventId = event.id) },
-        )
+        eventDao.upsertSyncStateAndEvent(eventEntity = event.toEntity(), eventSyncEntity = EventSyncEntity(id = event.id, syncType = syncType))
+        pictureDao.upsertPictures(deletedPictures = event.deletedPictures, pictures = event.pictures.map { it.toEntity(eventId = event.id) })
+        attendeeDao.upsertAttendees(eventId = event.id, attendees = event.attendees.map { it.toEntity(eventId = event.id) })
```
philipplackner, line 40: Could be done via a transaction, so that if one call fails, the other ones will not be executed. Works with `db.withTransaction { /* DB calls here */ }`

philipplackner (approved):

PR #69, opened 2023-04-20
Feature/sync with local data

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/data/remote/worker/SyncRemoteWithLocalData.kt
```diff
@@ -0,0 +1,107 @@
...
+
+    private var jobs = mutableListOf<Job>()
+
+    override suspend fun doWork(): Result {
+        return try {
+            supervisorScope {
+                remindersSync(coroutineScope = this)
+                tasksSync(coroutineScope = this)
+                eventsSync(coroutineScope = this)
+                jobs.forEach { it.join() }
```
MKiperszmid, line 37: instead of having a list of jobs shared across the whole class 
```
private var jobs = mutableListOf<Job>()
```

You could just:

```
val reminders = remindersSync(coroutineScope = this)
val tasks = tasksSync(coroutineScope = this)
val events = eventsSync(coroutineScope = this)
(reminders + tasks + events).forEach { it.join() }
```

app/src/main/java/com/artemissoftware/tasky/agenda/data/remote/worker/SyncRemoteWithLocalData.kt
```diff
@@ -0,0 +1,107 @@
...
+                    }
+                    SyncType.DELETE -> {
+                        deleteReminderAndSync(syncState.itemId)
+                    }
+                    else -> Unit
+                }
+            }
+        }
+
+        jobs.addAll(job)
```
MKiperszmid, line 63: And you would just return this job, instead of adding it to the list

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/data/remote/worker/SyncRemoteWithLocalData.kt
ArtemisSoftware: Refactored

app/src/main/java/com/artemissoftware/tasky/agenda/data/remote/worker/SyncRemoteWithLocalData.kt
ArtemisSoftware: Refactored

PR #71, opened 2023-04-24
Feature/join leave attendee

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/AttendeeRepositoryImpl.kt
```diff
@@ -24,4 +28,15 @@ class AttendeeRepositoryImpl(
...
     }
+
+    override suspend fun deleteAttendee(eventId: String): DataResponse<Unit> {
+        attendeeDao.upsertAttendeeSync(AttendeeSyncEntity(id = eventId, syncType = SyncType.DELETE))
+
+        return try {
+            agendaApiSource.deleteAttendee(eventId = eventId)
+            DataResponse.Success(Unit)
+        } catch (ex: TaskyNetworkException) {
+            DataResponse.Error(exception = ex)
```
MKiperszmid, line 39: What if you delete the attendee from db, but api fails?

You shouldnt display that agenda items

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/Attendee.kt
```diff
@@ -4,5 +4,5 @@ data class Attendee(
     val fullName: String,
     val id: String,
     val email: String,
-    val isGoing: Boolean
+    var isGoing: Boolean
```
MKiperszmid, line 7: Data classes should almost always have vals

Why do you need a var?

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/dashboard/AgendaViewModel.kt
```diff
@@ -159,7 +165,7 @@ class AgendaViewModel @Inject constructor(
                     sendUiEvent(UiEvent.Navigate(TaskDetailScreenDestination(taskId = item.itemId).route))
                 }
                 is AgendaItem.Event -> {
-                    sendUiEvent(UiEvent.Navigate(EventDetailScreenDestination(eventId = item.itemId, userId = _state.value.userId).route))
+                    sendUiEvent(UiEvent.Navigate(EventDetailScreenDestination(eventId = item.itemId, userId = _state.value.userId, userName = _state.value.userName).route))
```
MKiperszmid, line 168: Why do you need to pass the userid and the usrrname?

You could inject the SharedPreferences to the event detail viewmodel and get that info directly from there

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailScreen.kt
```diff
@@ -372,10 +389,15 @@ private fun LazyListScope.visitors(
     type: VisitorOptionType,
```
MKiperszmid, line 389: I'm not fully understanding the idea of this "visitors" function

Do you have a video showcasing it?

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/mappers/VisitorMapper.kt
```diff
@@ -0,0 +1,21 @@
...
+
+fun List<Attendee>.toNotGoingVisitors(): List<Visitor> {
+    return this.filter { !it.isGoing }.map { it.toVisitor(isEventCreator = false) }
+}
+
+fun List<Attendee>.toGoingVisitors(hostId: String): List<Visitor> {
+    val list = this.filter { it.isGoing }.map { it.toVisitor(isEventCreator = (it.id == hostId)) }.toMutableList()
+    list.sortBy { !it.isEventCreator }
+    return list.toList()
+}
```
MKiperszmid, line 21: Mappers should map objects from one type, to another

They shouldn't filter

This is better done on the Viewmodel

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailViewModel.kt
```diff
@@ -334,15 +357,30 @@ class EventDetailViewModel @Inject constructor(
                             pictures = item.pictures,
                             attendees = item.attendees,
                             hostId = item.hostId,
+                            goingVisitors = item.attendees.toGoingVisitors(hostId = item.hostId),
+                            notGoingVisitors = item.attendees.toNotGoingVisitors(),
```
MKiperszmid, line 361: A better way in my opinion, is to have a "filteredVisitors"

And simply apply a filter there, every time the visitorOption gets changed

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailViewModel.kt
```diff
@@ -334,15 +357,30 @@ class EventDetailViewModel @Inject constructor(
...
                         )
                     }
+                    safeLet(savedStateHandle.get<String>(USER_ID), savedStateHandle.get<String>(USER_NAME)) { userId, userName ->
+                        val isEventCreator = (item.hostId == userId)
+                        update {
+                            it.copy(
+                                userId = userId,
+                                isEventCreator = isEventCreator,
+                                creator = if (isEventCreator) { Visitor(attendee = Attendee(fullName = userName, id = userId, email = "", isGoing = true), isEventCreator = true) } else null,
```
MKiperszmid, line 370: No need for a "Creator" object tbh

You can simply have the filteredList as I mentioned, filled with Visitor

And have a true/false if it's the host

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailViewModel.kt
```diff
@@ -334,15 +357,30 @@ class EventDetailViewModel @Inject constructor(
...
                 }
             }
         } ?: run {
-            savedStateHandle.get<String>(USER_ID)?.let { userId ->
+            safeLet(savedStateHandle.get<String>(USER_ID), savedStateHandle.get<String>(USER_NAME)) { userId, userName ->
                 update {
                     it.copy(
                         hostId = userId,
+                        hostName = userName,
+                        creator = Visitor(attendee = Attendee(fullName = userName, id = userId, email = "", isGoing = true), isEventCreator = true),
```
MKiperszmid, line 383: Will you ever use the "email" in the visitor item? If not, what's the use of it?

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/AgendaItemType.kt
```diff
@@ -10,9 +10,10 @@ sealed class AgendaItemType(
...
-){
-    class Reminder(): AgendaItemType(color = Light2, borderColor = Gray, title = R.string.reminder)
-    class Task(color: Color = Green): AgendaItemType(color = color, borderColor = color, title = R.string.task, generalTextColor = White, secondaryTextColor = White)
-    class Event(color: Color = LightGreen): AgendaItemType(color = color, borderColor = color, title = R.string.event)
+    val secondaryTextColor: Color = DarkGray,
+    val bulletColor: Color = Black,
+) {
+    class Reminder() : AgendaItemType(color = Light2, borderColor = Gray, title = R.string.reminder)
+    class Task(color: Color = Green) : AgendaItemType(color = color, borderColor = color, title = R.string.task, generalTextColor = White, secondaryTextColor = White)
+    class Event(color: Color = LightGreen) : AgendaItemType(color = color, borderColor = color, title = R.string.event)
```
philipplackner, line 18: The colors won't change, so you can hardcode these in the constructor of `AgendaItemType` and all these single options can be `objects` instead since they don't have a constructor

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/AttendeeRepositoryImpl.kt
philipplackner: Yes, the `upsertAttendeeSync` call should be executed after the attendee has been successfully deleted on the API

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/Attendee.kt
philipplackner: Reason: Not vulnerable to race conditions that way

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/dashboard/AgendaViewModel.kt
philipplackner: Especially because the local storage serves as a single source of truth

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailScreen.kt
```diff
@@ -134,7 +141,7 @@ private fun EventDetailScreenContent(
                 ),
                 toolbarActions = { color ->
 
-                    if (state.isEditing) {
+                    if (state.isEditionOccurring()) {
```
philipplackner, line 144: I find the old name already says everything in less characters 🤔

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailScreen.kt
philipplackner: I assume he uses a single LazyColumn for the whole event detail screen and wants to place the single visitor items in that outer Column while still keeping the lazy loading behavior. Totally fine to do it this way

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailViewModel.kt
```diff
@@ -116,12 +121,29 @@ class EventDetailViewModel @Inject constructor(
...
+                setAttendeeGoingStatus(isGoing = false)
+            }
             else -> Unit
         }
     }
 
+    private fun setAttendeeGoingStatus(isGoing: Boolean) = with(_state) {
+        update {
+            val list = value.attendees.toMutableList()
+            list.find { item -> item.id == value.userId }?.isGoing = isGoing
```
philipplackner, line 137: With our suggestion to keep `isGoing` a val, better do it with `map`:
```
val newList = oldList.map {
    if(it.id == value.userId) it.copy(isGoing = isGoing) else it
}
```

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailViewModel.kt
philipplackner: You then still need to maintain the full list of visitors, otherwise the filter will apply on the filtered list. I personally left the full list in the state and filtered in the UI inside `remember(attendees)` blocks

philipplackner (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailViewModel.kt
MKiperszmid: Yes yes, I meant having an "attendees" list, and then a filtered one, where he can then apply a filter to the attendees based on the visitorOption

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/AttendeeRepositoryImpl.kt
ArtemisSoftware: I think there is confusion here. 
With the `upsertAttendeeSync ` I am only creating a reference that the attendee was deleted from the event.

If the api call to remove the attendee  from the event fails this reference will enable me that in a future syncronization with the api this reference will be used.

Its the same logic I am using with reminders, tasks and events

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/AttendeeRepositoryImpl.kt
MKiperszmid: So when are you removing them from the DB?

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailScreen.kt
```diff
@@ -161,26 +178,23 @@ private fun EventDetailScreenContent(
...
                             item {
                                 AssignmentHeader(
-                                    agendaItemType = AgendaItemType.Task(),
+                                    agendaItemType = AgendaItemType.Event,
                                     title = state.title,
                                     modifier = Modifier
                                         .fillMaxWidth()
                                         .padding(horizontal = 16.dp),
-                                    isEditing = state.isEditing,
+                                    isEditing = state.isEditing && state.isEventCreator,
```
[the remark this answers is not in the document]
ArtemisSoftware: When an attendee is on the EventDetailScreen he can only edit the notification.

This was the only way I found to add this logic.
I am not sure I can achieve this with what you are suggesting

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailScreen.kt
```diff
@@ -118,6 +125,16 @@ private fun EventDetailScreenContent(
         },
     )
 
+    val visitorsGoing = remember(key1 = state.attendees) {
+        val list = state.attendees.filter { it.isGoing }.map { it.toVisitor(isEventCreator = (it.id == state.hostId)) }.toMutableList()
+        list.sortBy { !it.isEventCreator }
```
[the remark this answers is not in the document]
MKiperszmid: Ahhh gotcha, sounds good

app/src/main/java/com/artemissoftware/tasky/agenda/presentation/detail/eventdetail/EventDetailState.kt
```diff
@@ -25,18 +24,6 @@ data class EventDetailState(
...
-
-    fun getNotGoingVisitors(): List<Visitor> {
-        return attendees.filter { !it.isGoing }.map { attendee -> Visitor(attendee = attendee, isEventCreator = isEventCreator(attendee.id)) }
-    }
-
-    private fun isEventCreator(attendeeId: String) = (attendeeId == hostId)
-
-    fun isGoing() = attendees.find { it.id == userId }?.isGoing ?: false
-}
+    val hostName: String = "",
```
MKiperszmid, line 27: Why do you need the host´s name tho?

MKiperszmid (approved):

app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/AttendeeRepositoryImpl.kt
ArtemisSoftware: On the DeleteAttendeeUseCase.

PR #72, opened 2023-04-25
Feature/logout observer

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/data/repositories/AgendaRepositoryImpl.kt
```diff
@@ -61,4 +61,14 @@ class AgendaRepositoryImpl(
             alarmScheduler.cancel(id)
         }
     }
+
+    override suspend fun deleteLocalAgenda() {
+        val deletedRemindersId = reminderDao.deleteAll()
+        val deletedTasksId = taskDao.deleteAll()
+        val deletedEventsId = eventDao.deleteAll()
```
MKiperszmid, line 68: You´d also need to delete the attendees and everything else, right?

core/src/main/java/com/artemissoftware/core/util/interceptors/JwtInterceptor.kt
```diff
@@ -10,24 +12,34 @@ import okhttp3.Response
 import retrofit2.Invocation
 import javax.inject.Inject
 
+private val _logOutState = MutableSharedFlow<Unit>(replay = 3)
+val logOutState = _logOutState.asSharedFlow()
```
MKiperszmid, line 16: Curious

Why did you go with a sharedflow instead of a stateflow?

MKiperszmid (approved):

PR #79, opened 2023-05-01
Feature/time needle

ArtemisSoftware:


app/src/main/java/com/artemissoftware/tasky/agenda/presentation/dashboard/AgendaViewModel.kt
```diff
@@ -204,10 +200,39 @@ class AgendaViewModel @Inject constructor(
...
+                    _state.update {
+                        it.copy(
+                            needlePosition = getNeedlePosition(result),
+                        )
+                    }
+                    delay(60.seconds)
+                }
+            }
+        }
+    }
```
philipplackner, line 224: Supervisor scope isn't necessary here and doesn't do much. You can just do
```
needleJob = viewModelScope.launch { ... }
```

Also, one way to pimp this a bit would be to check at the beginning how much time it is until the next full minute is reached. E.g. if it's 15:00:45, you'd delay the first iteration for only 15s and from then on for 60s. That way it'll always be updated at each starting minute which is probably what a user would expect.
But not a requirement, I didn't do this either.

philipplackner (approved):

PR #80, opened 2023-05-02
Feature/deep link

ArtemisSoftware:


.idea/androidTestResultsUserPreferences.xml
```diff
@@ -0,0 +1,23 @@
...
+                  <entry key="samsung&#10; SM-N950F" value="120" />
+                </map>
+              </option>
+            </AndroidTestResultsTableState>
+          </value>
+        </entry>
+      </map>
+    </option>
+  </component>
+</project>
```
MKiperszmid, line 23: You are uploading files that are in the .idea folder
That´s not really useful tbh, since if a team member downloads the project, this file will override their android studio configurations

I recommend adding it to git ignore by adding:

.idea

app/src/main/java/com/artemissoftware/tasky/agenda/domain/models/AgendaItemType.kt
```diff
@@ -1,20 +1,38 @@
...
-    EVENT(UiText.StringResource(R.string.event)),
-    REMINDER(UiText.StringResource(R.string.reminder)),
-    TASK(UiText.StringResource(R.string.task));
+sealed class AgendaItemType(
+    @StringRes val title: Int,
+    val color: Color,
+    val borderColor: Color,
+    val generalTextColor: Color = Black,
+    val secondaryTextColor: Color = DarkGray,
+    val bulletColor: Color = Black,
```
MKiperszmid, line 23: You are in the domain module, but are adding stuff that´s only used on the UI

app/src/main/java/com/artemissoftware/tasky/agenda/data/alarm/AlarmReceiver.kt
```diff
@@ -17,23 +18,22 @@ class AlarmReceiver : BroadcastReceiver() {
             extras.getString(ID)?.let { id ->
                 val title = extras.getString(TITLE).orEmpty()
                 val body = extras.getString(BODY).orEmpty()
+                val deeplink = extras.getString(LINK, "https://tasky.com")
```
MKiperszmid, line 21: The "https://tasky.com" could be a constant, since you are using it in many places

app/src/main/java/com/artemissoftware/tasky/agenda/data/alarm/AlarmSchedulerImpl.kt
```diff
@@ -55,9 +56,10 @@ class AlarmSchedulerImpl @Inject constructor(
     private fun getIntent(item: AgendaItem): Intent {
         val intent = Intent(context, AlarmReceiver::class.java)
         val bundle = Bundle().apply {
-            putString(TITLE, AgendaItemType.convertAgendaItem(item).name)
+            putInt(TITLE, AgendaItemType.convertAgendaItem(item).title)
```
MKiperszmid, line 59: The title is an int because it´s a string resource id?
I would name it "titleId" then, to make it a bit more easier to understand

MKiperszmid (approved):

.idea/androidTestResultsUserPreferences.xml
ArtemisSoftware: Added

PR #82, opened 2023-05-08
Feature/delete user

ArtemisSoftware:


core/src/main/java/com/artemissoftware/core/data/repositories/UserRepositoryImpl.kt
```diff
@@ -0,0 +1,46 @@
...
+        context.userStore.updateData {
+            User()
+        }
+    }
+
+    override suspend fun deleteAllUserData(): List<String> {
+        val deletedRemindersId = reminderDao.deleteAll()
+        val deletedTasksId = taskDao.deleteAll()
+        val deletedEventsId = eventDao.deleteAll()
+        return (deletedRemindersId + deletedTasksId + deletedEventsId)
```
MKiperszmid, line 44: Why do you need to return the ids tho?

core/src/main/java/com/artemissoftware/core/domain/usecase/DeleteAllUserDataUseCase.kt
```diff
@@ -0,0 +1,15 @@
...
+import com.artemissoftware.core.domain.alarm.AlarmScheduler
+import com.artemissoftware.core.domain.repositories.UserRepository
+import javax.inject.Inject
+
+class DeleteAllUserDataUseCase @Inject constructor(
+    private val userRepository: UserRepository,
+    private val alarmScheduler: AlarmScheduler,
+) {
+    suspend operator fun invoke() {
+        val ids = userRepository.deleteAllUserData()
```
MKiperszmid, line 12: Ohh I just saw you are using them here. Disregard my previous comment

core/src/main/java/com/artemissoftware/core/util/interceptors/JwtInterceptor.kt
```diff
@@ -38,6 +42,7 @@ class JwtInterceptor @Inject constructor(private val getUserUseCase: GetUserUseC
             if (response.code == 401) {
                 runBlocking {
                     _logOutState.emit(Unit)
+                    deleteAllUserDataUseCase.invoke()
```
MKiperszmid, line 45: Personally instead of calling invoke, I like to just:
```
deleteAllUserDataUseCase()
```

MKiperszmid (approved):

